PR #1, opened 2024-08-19
Mevblocker fee query

fleupold:
Should contain the main logic of https://github.com/cowprotocol/mev-blocker-dune-queries/pull/3/files#diff-00f13f97bd256af9bcc645bc1431b2c41632d594cd409937a22b2dd61af640ee with a few small changes

- Instead of grouping by month, group by block (so the query can be reused in different timeframes and for individual block debugging)
- Remove duplicate column names
- Include `backrun_tip` in fee computation (I believe this was a bug)
- Change range queries on block numbers and pre-filter relevant mevblocker bundles (this speeds up the query significantly)

Query has been linted using sqlfluff (will set this up as a standard for this repo going forward).

### DIFF
--- a/mevblocker/fees/.sqlfluff
+++ b/mevblocker/fees/.sqlfluff
@@ -1,0 +1,3 @@
+[sqlfluff:templater:jinja:context]
+start='2024-08-01 12:00'
+end='2024-08-02 12:00'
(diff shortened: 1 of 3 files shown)


olgafetisova:
Instead of grouping by month, group by block (so the query can be reused in different timeframes and for individual block debugging) - a very good approach, no performance issues was spotted here.

Pre-filtering tables before joining - good speed improvement.

The main difference was noticed due to backrun_tip in fee computation that was indeed calculated but omitted from the previous final value. Otherwise, the values match.

olgafetisova (approved):

PR #3, opened 2024-08-20
[MEVBlocker] Add builder info

fleupold:
Combined information of https://github.com/cowprotocol/mev-blocker-dune-queries/pull/3/files#diff-925cc26fdd0ed2249ae4c0ca85744420f8bb3332673294eae682e5d10acbbef4 and https://github.com/cowprotocol/mev-blocker-dune-queries/pull/3/files#diff-1f097840c4dc071ffdc139293d027efd5e49cc86d418d9f1f0abbc6b96c303a0

containing all relevant subscription information for the MEV Blocker builders


olgafetisova:
Any reason to omit `bond_provider` part as here https://github.com/cowprotocol/mev-blocker-dune-queries/blob/9ddfc5210842a3ccbfa9b8369f893e7777f45f3e/queries/block_builders___3665113.sql ?

fleupold:
Happy to add it, but it was set as null everywhere. Do we have the values that should be inserted there somewhere? What is bond provider used for?

olgafetisova:
> Happy to add it, but it was set as null everywhere. Do we have the values that should be inserted there somewhere? What is bond provider used for?

Yes, it has values for Flashbots at this stage due to the way they deposited ETH.
This is to support this PR https://github.com/cowdao-grants/mb-billing/pull/29#issuecomment-2270908544

fleupold:
Ok, so only relevant for flashbots, and the value should be `0xa489faf6e337d997b8a23e2b6f3a8880b1b61e19`

PR #6, opened 2024-08-25
Add query for payment status

fleupold:
Last query (basically taken from https://dune.com/queries/3742749) computes the payment status for each builder (given the latest billing).

We might want to change this query in the future to account for multiple bill transactions (e.g. in case of manual adjustment).

### DIFF
--- a/mevblocker/fees/payment_status_4016152.sql
+++ b/mevblocker/fees/payment_status_4016152.sql
@@ -1,0 +1,45 @@
+with
+latest_bill_date as (
+    select max(call_block_time) as bill_date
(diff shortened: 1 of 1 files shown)


fleupold:
Yes, but the payer is not taken from the tranaction sender but instead the argument in the event e.g. this one: https://etherscan.io/tx/0x768021dd497ee7ce7ebd359ff3ccff1b9c6b79fcad813039eb8043b081abb00f#eventlog

You have to pay for the correct account.

olgafetisova (approved):

PR #8, opened 2024-09-02
Add mainnet solver rewards dashboard and update with service fee

harisang:
This PR adds the following to the repo:
- [Query 2510345](https://dune.com/queries/2510345), which is the main query showing in the solver rewards [dashboard](https://dune.com/cowprotocol/cow-solver-rewards?StartTime=2024-08-20+00%3A00%3A00&EndTime=2024-08-27+00%3A00%3A00).
- [Query 4017925](https://dune.com/queries/4017925), which computes whether a solver needs to pay the CoW DAO bonding pool service fee or not.

Note that query 2510345 has some proposed changes, regarding the implementation of the service fee and the reduction in the rewards being transferred to solvers. The changes are between lines 746-811.

So the plan is to get the PR approved, and before merging actually edit query 2510345 to ensure its code is identical to the code in this PR.

### DIFF
--- a/mainnet_solver_rewards/service_fee_query_4017925.sql
+++ b/mainnet_solver_rewards/service_fee_query_4017925.sql
@@ -1,0 +1,175 @@
+WITH
+bonding_pools (pool, name, initial_funder) AS (
+  SELECT from_hex(pool), name, from_hex(funder) FROM (
+    VALUES {{bondingpooldata}}
+  ) AS _ (pool, name, funder)
+),
+
+first_event_after_timestamp AS (
+  SELECT MAX(number) FROM ethereum.blocks
+  WHERE time > CAST('2024-08-20 00:00:00' AS timestamp) -- CIP-48 starts bonding pool timer at midnight UTC on 20/08/24
+),
+
+initial_vouches AS (
+  SELECT RANK() OVER (
+      PARTITION BY solver, bondingPool, sender
+      ORDER BY evt_block_number ASC, evt_index ASC 
+    ) AS rk,
+    evt_block_number,
+    evt_index,
+    solver,
+    cowRewardTarget,
+    bondingPool,
+    sender,
+    True AS active
+  FROM cow_protocol_ethereum.VouchRegister_evt_Vouch
+  WHERE evt_block_number <= (SELECT * FROM first_event_after_timestamp)
+    AND bondingPool IN (SELECT pool FROM bonding_pools)
+    AND sender IN (SELECT initial_funder FROM bonding_pools)
+),
+
+joined_on_data AS (
+  SELECT
+    iv.solver,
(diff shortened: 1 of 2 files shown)


mainnet_solver_rewards/service_fee_query_4017925.sql
```diff
@@ -0,0 +1,175 @@
...
+),
+
+named_results AS (
+  SELECT
+    jd.solver,
+    CONCAT(environment, '-', s.name) AS solver_name,
+    jd.pool_name,
+    jd.pool,
+    jd.joined_on,
+    date_diff('day', date(jd.joined_on), date(NOW())) AS days_in_pool
```
fhenneke, line 117: `NOW()` should probably be changed to a `cutoff_time` parameter which would chosen as `start_time`.

mainnet_solver_rewards/service_fee_query_4017925.sql
```diff
@@ -0,0 +1,175 @@
...
+    VALUES {{bondingpooldata}}
+  ) AS _ (pool, name, funder)
+),
+
+first_event_after_timestamp AS (
+  SELECT MAX(number) FROM ethereum.blocks
+  WHERE time > CAST('2024-08-20 00:00:00' AS timestamp) -- CIP-48 starts bonding pool timer at midnight UTC on 20/08/24
+),
+
+initial_vouches AS (
```
fhenneke, line 13: The part on vouches might just reuse the vouches query.

mainnet_solver_rewards/service_fee_query_4017925.sql
fhenneke, line 1: That query does not look much simpler that just hard-coding all solver addresses, their starting date, and their status (e.g. exempt for 6 month, 9 month, indefinitely).

The query might help finding those dates.

mainnet_solver_rewards/service_fee_query_4017925.sql
```diff
@@ -0,0 +1,175 @@
+WITH
+bonding_pools (pool, name, initial_funder) AS (
+  SELECT from_hex(pool), name, from_hex(funder) FROM (
```
fhenneke, line 3: (this query uses a different style (keywords in all caps) than the other query)

fhenneke (changes requested):
This is a good opportunity to clean up this query (and the respective solver-rewards query).

As a minimum, the external queries referenced in this query should be called instead of their code being copied.

There is much room for other improvements, but some of that (e.g. using the actual solver-rewards query for the dashboard or using query 4021306 from coincidence of want for slippage) could be addressed later.

fhenneke (approved):

harisang:
Closing as a lot of changes have happened in between so will reopen a PR with just the main dashboard there.

PR #9, opened 2024-09-03
Add mainnet service fee tracker query (id 4017925)

harisang:
This PR simply adds this query for tracking purposes
https://dune.com/queries/4017925

### DIFF
--- a/cowprotocol/accounting/rewards/mainnet/service_fee_query_4017925.sql
+++ b/cowprotocol/accounting/rewards/mainnet/service_fee_query_4017925.sql
@@ -26,7 +26,7 @@ initial_vouches AS (
         cowRewardTarget,
         bondingPool,
         sender,
-        True AS active,
+        TRUE AS active,
         rank() OVER (
             PARTITION BY
                 solver,
@@ -64,7 +64,7 @@ joined_on_data AS (
         iv.evt_block_number,
         iv.evt_index,
         iv.rk,
-        True AS active
+        TRUE AS active
     FROM
         initial_vouches AS iv
     WHERE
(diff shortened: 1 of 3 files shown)


fleupold (commented):
I think this folder structure is not furture proof. I think this should go under cowprotocol (cf. @fhenneke's PR) -> accounting -> rewards -> mainnet

cowprotocol/accounting/rewards/mainnet/service_fee_query_4017925.sql
```diff
@@ -0,0 +1,253 @@
...
+
+bonding_pools (name, pool, initial_funder) AS (
+    SELECT
+        name,
+        from_hex(pool),
+        from_hex(funder)
+    FROM (
+        VALUES {{BondingPoolData}}
+    ) AS _ (name, pool,funder)
+),
```
bram-vdberg, line 11: I think we can fix the linting errors and also set the bonding pools as constants if we hardcode them like this:

```
WITH
  bonding_pools (pool, name, initial_funder) AS (
    SELECT
      from_hex('0x8353713b6D2F728Ed763a04B886B16aAD2b16eBD') AS pool,
      'Gnosis' AS name,
      from_hex('0x6c642cafcbd9d8383250bb25f67ae409147f78b2') AS initial_funder
    UNION ALL
    SELECT
      from_hex('0x5d4020b9261F01B6f8a45db929704b0Ad6F5e9E6') AS pool,
      'CoW Services' AS name,
      from_hex('0x423cec87f19f0778f549846e0801ee267a917935') AS initial_funder
  ),
```

bram-vdberg (approved):

fleupold:
> I have also edited the query on Dune so that it is identical to the latest commit in this PR

Note that this should not be needed, merging the PR will automatically override what is stored in Dune

PR #10, opened 2024-09-04
Fix capitalization in service fee query (id 4017925)

harisang:
Trivial PR that lowercases the service fee query and uses snake case throughout for column names (4017925)

### DIFF
--- a/cowprotocol/accounting/rewards/mainnet/service_fee_query_4017925.sql
+++ b/cowprotocol/accounting/rewards/mainnet/service_fee_query_4017925.sql
@@ -124,7 +124,7 @@ latest_vouches as (
                 evt_block_number,
                 evt_index,
                 solver,
-                NULL as cow_reward_target, -- Invalidation does not have a reward target
+                null as cow_reward_target, -- Invalidation does not have a reward target
                 bonding_pool,
                 sender,
                 'InvalidateVouch' as event_type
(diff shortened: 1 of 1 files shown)


harisang:
@fhenneke In case you have other stylistic suggestions, this might be the PR to apply them. E.g, column names. Should we use underscores and lowercase letters instead of names such as `bondingPool` ?

harisang:
> @fhenneke In case you have other stylistic suggestions, this might be the PR to apply them. E.g, column names. Should we use underscores and lowercase letters instead of names such as `bondingPool` ?

Actually pushed a change to snake case throughout

harisang:
> LGTM, what's the reason for removing capitalization? I thought capitalizing SQL specific words was the standard syntax.

Yeah that's a valid point. I guess currently we are inconsistent and i decided to propose going with lower case. Whatever we choose, it would be nice to stick to it.

Do you have any preference @fhenneke ?

fhenneke:
I do not really have an opinion on capitalization. The most important part to me would be would be that it is consistens (first within a query and then across queries).

If we cafe about a specific styling, we can add it to the linting check, e.g. via
```
[sqlfluff:rules:capitalisation.keywords]
capitalisation_policy = lower
```

harisang:
Ok, i would suggest we keep this as is then (lower case) and try to be consistent throughout. Merging so that we can finally start fixing the query

PR #13, opened 2024-09-05
Fix CIP-48 query (id 4017925)

bram-vdberg:
The recent changes to the CIP-48 query seem to have introduced an error. This is because the `cow_protocol_ethereum.VouchRegister_evt_Vouch` table uses camel case for the `bondingPool` and `CowRewardTarget` columns. This PR reverts those columns in the CIP-48 query to a naming format that is compatible with the table.


harisang:
Oh good catch, and sorry for introducing this error!

harisang (approved):

Issue #12, opened 2024-09-05
Add Hardcoded Colocated Solver Addresses for CoW Bonding Pool


bram-vdberg:
(Part of #28)

**Background**
At the moment the smart contract team is working on a solution to keep track of colocated solver addresses within the CoW Bonding Pool. However, until this contract is ready, we should hard-code the relevant addresses and update them as needed.

**Proposed Solution**
- [ ] Update CMS to track the bonding pool that each solver is in
- [x] Update the [service_fee_query_4017925.sql](https://github.com/cowprotocol/dune-queries/blob/main/cowprotocol/accounting/rewards/mainnet/service_fee_query_4017925.sql) to hard-code the addresses of all colocated solvers on line 4-12 where we define the `bonding_pools` (mirroring the CMS).
- [ ] Create GitHub action to update this query based on the data in the CMS

bram-vdberg:
I think that sounds like the best approach actually. At the moment we don't seem to track the bonding pool information in the CMS so we would have to update that as well.

PR #14, opened 2024-09-05
Update query to use hardcoded values for colocated solvers

bram-vdberg:
This PR addresses issue #12 and updates the CIP-48 query to use hard-coded values for colocated drivers. 

This change also updates the logic for calculating the 3 month grace period for colocated solvers. Now, the logic only sets the grace period if they are manually marked as being in a reduced bonding pool.

Currently hardcoding all colocated solvers. Updates have been made to the CMS to allow us to get this information in a GitHub action, but since that's not live yet I'm adding these manually.

### DIFF
--- a/cowprotocol/accounting/rewards/mainnet/service_fee_query_4017925.sql
+++ b/cowprotocol/accounting/rewards/mainnet/service_fee_query_4017925.sql
@@ -6,8 +6,8 @@ colocated_solvers as (
         'prod-Barter' as solver_name,
         'Reduced-Bonding' as pool_name,
         from_hex('0xB6113c260aD0a8A086f1E31c5C92455252A53Fb8') as pool,
-        timestamp '2024-08-21 07:15:00' as joined_on,
-        from_hex('0xC7899Ff6A3aC2FF59261bD960A8C880DF06E1041') as solver
+        from_hex('0xC7899Ff6A3aC2FF59261bD960A8C880DF06E1041') as solver,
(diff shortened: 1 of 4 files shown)


cowprotocol/accounting/rewards/mainnet/service_fee_query_4017925.sql
```diff
@@ -1,5 +1,15 @@
 with
 
+-- Add colocated solvers here
+colocated_solvers as (
+    select
+        'prod-Barter' as solver_name,
```
harisang, line 6: note we also have barn-Barter and prod-Copium_Capital that are colocated

bram-vdberg: Thanks, I was hoping to try filling them automatically using the CMS. If I can't set that up on time before Monday I'll just add those manually here and set this PR as ready to review.

harisang:
I took the liberty to add an alternative query that is simpler, at least to me, to parse. It is definitely doing redundant things and recomputing things here and there, but i had a bit of a hard time parsing the original query, so this is why i decided to do this.

I am not suggesting we should use the new one, but would be good to have a look at both, proof-read them and then make a call

harisang:
One thing that needs to happen is that the original query should better invoke query 1541516 as black box, even if that means repeating some calculations. @bram-vdberg would be good to check how much this would simplify your query

harisang:
Merging now as we need to execute the payouts tomorrow. Will open an issue to simplify the query

PR #16, opened 2024-09-05
Create uni_v2_amm_lp_and_tvl_4047194

olgafetisova:
Uni v2 pool is found based on the cow amm pool composition (the user inputs cow amm pool and the query finds an equivalent for uni pool).

Lp token supply is calculated based on the in and out transfer amounts on the token (pool address).

Then the query calculates token reserves and finds the latest daily price to calculate TVL.


cow_amm/uni_v2_amm_lp_and_tvl_4047194
```diff
@@ -0,0 +1,94 @@
...
+
+, uniswap_pool as 
+    (select output_0 as pool from uniswap_v2_ethereum.Factory_call_getPair
+    where ((_0 in (select token_1_address from cow_amm_pool) and _1 in (select token_2_address from cow_amm_pool) )
+    or (_0 in (select token_2_address from cow_amm_pool) and _1 in (select token_1_address from cow_amm_pool)))
+    and not(output_0 = 0x0000000000000000000000000000000000000000)
+    limit 1
+    )
+    
+, get_lp_balance as (
```
harisang, line 16: I would maybe rename this a bit. If i understand correctly, this computes how many LP tokens were minted and burned on a given day, right? So maybe something very literal as `lp_token_mint_burn_per_day`

cow_amm/uni_v2_amm_lp_and_tvl_4047194
```diff
@@ -0,0 +1,94 @@
...
+        ,ROW_NUMBER() OVER (PARTITION BY date_trunc('day', evt_block_time) ORDER BY evt_block_time ASC ) AS sync_count_in_day
+    FROM uniswap_v2_ethereum.Pair_evt_Sync
+    WHERE contract_address in (select pool from uniswap_pool)
+    ORDER BY evt_block_time DESC
+)
+,pair_sync_last_in_day AS (
+    SELECT *, MAX(sync_count_in_day) OVER (PARTITION BY date_trunc('day', evt_block_time)) AS max_sync_count_in_day
+    FROM pair_sync_with_counts
+)
+,v2_pair_data AS (
```
harisang, line 41: Tbh it is not clear to me what info we are trying to recover in lines 41-51 that we don't already have.

cow_amm/uni_v2_amm_lp_and_tvl_4047194
```diff
@@ -0,0 +1,94 @@
...
+    FROM prices.usd_daily
+    WHERE blockchain in (select blockchain from cow_amm_pool)
+        AND contract_address IN (SELECT token_address FROM token_address_list)
+)
+
+SELECT 
+    r.contract_address
+    , r.token0
+    , r.token1
+    , month as day
```
harisang, line 83: Ha, this also looks a bit odd.

cow_amm/uni_v2_amm_lp_and_tvl_4047194
```diff
@@ -0,0 +1,94 @@
...
+    LEFT JOIN tokens.erc20 AS t0 ON t0.contract_address = pc.token0 AND t0.blockchain in (select blockchain from cow_amm_pool)
+    LEFT JOIN tokens.erc20 AS t1 ON t1.contract_address = pc.token1 AND t1.blockchain in (select blockchain from cow_amm_pool)
+    WHERE pc.pair in (select pool from uniswap_pool)
+)
+,token_address_list AS (
+    SELECT token0 AS token_address FROM v2_pair_data
+    UNION ALL
+    SELECT token1 AS token_address FROM v2_pair_data
+)
+,v2_reserves_by_day AS (
```
harisang, line 58: to compute this, we could also do something similar to what you do for LP tokens, i.e., track the transfers in and out of the pool, right?

I wonder also whether there are cases of random transfers to the pool that are not proper liquidity injections (not sure why anyone would do that though), and whether in that case there is no Sync() event emitted.

cow_amm/uni_v2_amm_lp_and_tvl_4047194
```diff
@@ -0,0 +1,94 @@
...
+    WHERE blockchain in (select blockchain from cow_amm_pool)
+        AND contract_address IN (SELECT token_address FROM token_address_list)
+)
+
+SELECT 
+    r.contract_address
+    , r.token0
+    , r.token1
+    , month as day
+    , total_lp
```
harisang, line 84: yeah the names here, `total_lp` and `lp_supply` are a bit confusing to me.

cow_amm/uni_v2_amm_lp_and_tvl_4047194
fhenneke, line 1: Could you rename to have an ".sql" at the end? There is no syntax highlighting and not linting for the current file name.

cow_amm/uni_v2_amm_lp_and_tvl_4047194
```diff
@@ -0,0 +1,94 @@
+WITH 
+cow_amm_pool as (
+select blockchain, address, token_1_address, token_2_address
+from query_3959044 q
```
fhenneke, line 4: That query should also be added to the repo at some point.

cow_amm/uni_v2_amm_lp_and_tvl_4047194
```diff
@@ -0,0 +1,94 @@
...
+where address={{cow_amm_pool}}
```
fleupold, line 5: I wonder if this is the right input for the comparison? This way they can only be used for CoW AMM pools (where in theory profitability could be computed for any AMM regardless whether there exists a CoW AMM)\

Maybe we should have all queries be parametered on token1, token2 and then put the search for CoW AMMs in the top query that calls the subqueries.

cow_amm/uni_v2_amm_lp_and_tvl_4047194
```diff
@@ -0,0 +1,94 @@
...
+)
+
+,prices AS (
+    SELECT day, contract_address,symbol, price_close as price
+    FROM prices.usd_daily
+    WHERE blockchain in (select blockchain from cow_amm_pool)
+        AND contract_address IN (SELECT token_address FROM token_address_list)
+)
+
+SELECT
```
fleupold, line 79: This needs to have the same columns as the cow table so they can be put together in the top level query.

cow_amm/uni_v2_amm_lp_and_tvl_4047194
```diff
@@ -0,0 +1,94 @@
...
+
+, pair_sync_with_counts AS (
+    SELECT *
+        ,ROW_NUMBER() OVER (PARTITION BY date_trunc('day', evt_block_time) ORDER BY evt_block_time ASC ) AS sync_count_in_day
+    FROM uniswap_v2_ethereum.Pair_evt_Sync
+    WHERE contract_address in (select pool from uniswap_pool)
+    ORDER BY evt_block_time DESC
+)
+,pair_sync_last_in_day AS (
+    SELECT *, MAX(sync_count_in_day) OVER (PARTITION BY date_trunc('day', evt_block_time)) AS max_sync_count_in_day
```
fleupold, line 38: Can't you simply order DESC when partition above and then use rank 1?

cow_amm/uni_v2_amm_lp_and_tvl_4047194
```diff
@@ -0,0 +1,94 @@
...
+
+, total_lp as (
+select day, lp_supply, sum(lp_supply) over (order by day) as total_lp
+from get_lp_balance
+)
+
+, pair_sync_with_counts AS (
+    SELECT *
+        ,ROW_NUMBER() OVER (PARTITION BY date_trunc('day', evt_block_time) ORDER BY evt_block_time ASC ) AS sync_count_in_day
+    FROM uniswap_v2_ethereum.Pair_evt_Sync
```
fleupold, line 33: Very good idea to use this event!

cow_amm/uni_v2_amm_lp_and_tvl_4047194
```diff
@@ -0,0 +1,94 @@
...
+        , CAST(s.reserve1 AS DOUBLE)/POW(10,t1_decimals) as token_2_reserve
+    FROM pair_sync_last_in_day AS s
+    LEFT JOIN v2_pair_data AS pc ON pc.pair = s.contract_address
+    WHERE sync_count_in_day = max_sync_count_in_day
+    ORDER BY month DESC
+)
+
+,prices AS (
+    SELECT day, contract_address,symbol, price_close as price
+    FROM prices.usd_daily
```
fleupold, line 74: you can get the decimals from here.

cow_amm/uni_v2_amm_lp_and_tvl_4047194
```diff
@@ -0,0 +1,94 @@
...
+    SELECT token0 AS token_address FROM v2_pair_data
+    UNION ALL
+    SELECT token1 AS token_address FROM v2_pair_data
+)
+,v2_reserves_by_day AS (
+    SELECT
+        s.contract_address
+        ,pc.token0
+        ,pc.token1
+        ,date_trunc('day',s.evt_block_time) AS month
```
fleupold, line 63: This is very misleading naming, calling a day month

fleupold (changes requested):
Likewise, please provide good comments on how and what is being done. I think this query can be simplified quite a bit and should have the same output format as the cow one

cow_amm/uni_v2_amm_lp_and_tvl_4047194
olgafetisova: I thought of adding token_1 and token_2 parameters, the issue see here is how to know which token is 1 and which one is 2? I think there is an alphabetical order but does it always stand? Should we allow token symbol inputs or raw tokens (more accurate)?

fleupold: Def. addresses and not symbols. As to ordering, if we cannot make it ordering agnostic you can also follow @fhenneke's suggestion of using the uniswap pool as an argument. Note, that this query is not designed to be used by itself but instead will be embedded in a simple visualisation query which can make sure the input parameter is well defined (e.g. by choosing it from a hard coded list or the like).

cow_amm/uni_v2_amm_lp_and_tvl_4047194
```diff
@@ -0,0 +1,94 @@
...
+select date(evt_block_time) as day, 
+sum(case when "from" = 0x0000000000000000000000000000000000000000 THEN (value/1e18) ELSE -(value/1e18) END) as lp_supply
+from erc20_ethereum.evt_transfer
+where contract_address in (select pool from uniswap_pool)
+and ("from" = 0x0000000000000000000000000000000000000000 or "to" = 0x0000000000000000000000000000000000000000)
+group by 1
+)
+
+, total_lp as (
+select day, lp_supply, sum(lp_supply) over (order by day) as total_lp
```
fleupold: What happens here if a day passes without any mint or burn? We would have a gap in the data even though trading may have happened?

cow_amm/uni_v2_amm_lp_and_tvl_4047194
```diff
@@ -0,0 +1,94 @@
...
+    (select output_0 as pool from uniswap_v2_ethereum.Factory_call_getPair
+    where ((_0 in (select token_1_address from cow_amm_pool) and _1 in (select token_2_address from cow_amm_pool) )
+    or (_0 in (select token_2_address from cow_amm_pool) and _1 in (select token_1_address from cow_amm_pool)))
+    and not(output_0 = 0x0000000000000000000000000000000000000000)
+    limit 1
+    )
+    
+, get_lp_balance as (
+select date(evt_block_time) as day, 
+sum(case when "from" = 0x0000000000000000000000000000000000000000 THEN (value/1e18) ELSE -(value/1e18) END) as lp_supply
```
fleupold: Dividing by 1e18 here leads to float arithmetic rounding errors downstream, which make it hard to verify the output data is exactly correct (by checking with on chain values)

cow_amm/uni_v2_amm_lp_and_tvl_4047194
fleupold: Thinking about this more, I believe the input should be two tokens. This way each "$10k growth" chart is independent of cow amms (e.g. we could still compare uni, balancer and a counterfactual rebalancing portfolio even if no CoW AMM existed).

fleupold (approved):

fleupold:
@olgafetisova can you please take a look at the changes I made and incorporate those into #17?

cow_amm/uni_v2_amm_lp_and_tvl_4047194.sql
```diff
@@ -0,0 +1,163 @@
...
+            date_range.day,
+            total_lp,
+            rank() over (partition by (date_range.day) order by lp.day desc) as latest
+        from date_range
+        inner join lp_total_supply_incomplete as lp
+            on date_range.day >= lp.day
+            -- performance optimisation: this assumes one week prior to start there was at least one lp supply change event
+            and lp.day >= (timestamp '{{start}}' - interval '7' day)
+    )
+    where latest = 1
```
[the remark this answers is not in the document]
fleupold: > is it supposed to take the previous available value if the data is missing for a certain date?

Yes exactly. Can you elaborate why you think this may not be working? 

Wouldn't we in your suggestion still miss a record if there are missing values for two days in a row (which is the case for less frequently traded pairs)? Cf minimal example query: https://dune.com/queries/4055768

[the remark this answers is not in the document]
olgafetisova: Previous approach did not work for me initially, but after rerunning it again it seems to be fine.
Initially, if I selected the date with the missing values, it would not find any corresponding previous values, but I cannot replicate it anymore, so all fine from my size now. Adjusted my query as well.

PR #17, opened 2024-09-05
Create cow_amm_lp_and_tvl_4047078.sql

olgafetisova:
The query find the relevant cow amm pool, get lp token balance based on the incoming and outgoing ethereum transfers. 

Then we use the existing materialized table (result_balancer_cow_amm_base_query_v_2) to calculate TVL for the same pool.


cow_amm/cow_amm_lp_and_tvl_4047078
```diff
@@ -0,0 +1,48 @@
+with
```
fleupold, line 1: Can you comment the high level query file (documenting required parameters) and even the subqueries (I think the other files in the repo can serve as an example) so that it's clear what and how this query is doing things?

cow_amm/cow_amm_lp_and_tvl_4047078
```diff
@@ -0,0 +1,48 @@
+with 
+cow_amm_pool as (
```
fleupold, line 2: This subquery seems unused

cow_amm/cow_amm_lp_and_tvl_4047078
```diff
@@ -0,0 +1,48 @@
...
+
+select 
+  t1.day, 
+  total_tvl, 
+  total_lp, 
+  lp_supply,
+  total_tvl/total_lp as lp_token_price
+from total_tvl t1
+join total_lp t2 on t1.day=t2.day
+where t1.day>=date'2024-07-30'
```
fleupold, line 47: This should be a parameter (start date of the analysis)

cow_amm/cow_amm_lp_and_tvl_4047078
```diff
@@ -0,0 +1,48 @@
...
+, total_lp as (
+select day, lp_supply, sum(lp_supply) over (order by day) as total_lp
+from get_lp_balance
+)
+
+, get_tvl as (
+select 
+  date(time) as day, 
+  time, 
+  blockchain,
```
fleupold, line 25: nit: These two seem unused

cow_amm/cow_amm_lp_and_tvl_4047078
```diff
@@ -0,0 +1,48 @@
...
+
+, get_tvl as (
+select 
+  date(time) as day, 
+  time, 
+  blockchain, 
+  cow_amm_address, 
+  token_1_balance_usd+token_2_balance_usd as tvl,
+  RANK() OVER (PARTITION BY date(time) ORDER BY time DESC) AS tvl_rank
+from  dune.cowprotocol.result_balancer_cow_amm_base_query_v_2
```
fleupold, line 29: Are we sure we want to rely on this materialized view for TVL? Is it updating reliably now (last time I checked it was stuck many days in the past)? 

Have you tried using something like
```sql
select day, balance
from tokens_ethereum.balances_daily
where address = {{ cow_amm }}
and token_address = {{ token }}
and blockchain = 'ethereum'
```

to get the TVL?

cow_amm/cow_amm_lp_and_tvl_4047078
```diff
@@ -0,0 +1,48 @@
...
+select day,  blockchain, cow_amm_address, tvl as total_tvl
+from get_tvl
+where tvl_rank=1
+)
+
+select 
+  t1.day, 
+  total_tvl, 
+  total_lp, 
+  lp_supply,
```
fleupold, line 43: What is this? It's the supply delta per day, isn't it? I don't think this is relevant in the final table (I find current name is confusing given there is total_lp which is supposed the be the total_lp_supply).

fleupold (changes requested):
The file needs to end in `.sql` for it to be picked up. Also, please make sure formatting is correct (you can look at the other files in the rep). For this you likely need to edit the file in vscode or some other editor.

Also, given that the other folders are `mevblocker` and `cowprotocol`, the folder should probably be `cowamm`

cow_amm/cow_amm_lp_and_tvl_4047078
olgafetisova: Yes, I did check that table. It is way heavier and takes longer to compute the values (takes ~4mins to get those values), hence, I was trying to avoid it, but if there is more confidence I can update it. Ultimately it gives the same values. 

In the past few days, the table (dune.cowprotocol.result_balancer_cow_amm_base_query_v_2) didn't have any issues, but I did not find a way to check the historical view of the refreshes.

fleupold: Makes sense to not use `balances_daily` if it is too slow. I wonder if we can use a similar approach as for the LP tokens (summing up transfers of token1 and token2 in and out of this pool and multiply it with the price at the end of each day) to get the final balance efficiently without relying on the materialized view.

cow_amm/cow_amm_lp_and_tvl_4047078
fleupold: Yes, I think users should be able to chose a custom start date (we can set the default to the pool creation in the final visualisation)

cow_amm/cow_amm_lp_and_tvl_4047078
```diff
@@ -1,14 +1,34 @@
...
+        unnest(sequence(
+            date(timestamp '{{start}}') - INTERVAL '1' day,
+            date(now())
+        )) t (day) --noqa: AL01
+),
+
+-- Finds the CoW AMM pool address given tokens specified in query parameters (regardless of order) 
 cow_amm_pool as (
 select blockchain, address, token_1_address, token_2_address
 from query_3959044 q
```
fleupold: This query should also move under version_control. Can you add it in a separate PR please?

cow_amm/cow_amm_lp_and_tvl_4047078
```diff
@@ -0,0 +1,107 @@
...
+
+-- Finds the CoW AMM pool address given tokens specified in query parameters (regardless of order) 
+cow_amm_pool as (
+select blockchain, address, token_1_address, token_2_address
+from query_3959044 q
+where ((token_1_address = {{token_a}} and token_2_address = {{token_b}}) or (token_2_address = {{token_a}} and token_1_address = {{token_b}}))
+)
+
+-- per day lp token total supply changes of the CoW AMM pool by looking at burn/mint events
+, get_lp_balance as (
```
fleupold:
```suggestion
, lp_balance_delta as (
```

cow_amm/cow_amm_lp_and_tvl_4047078
```diff
@@ -0,0 +1,107 @@
...
+    select date(evt_block_time) as day, 
+    sum(case when "from" = 0x0000000000000000000000000000000000000000 THEN (value/1e18) ELSE -(value/1e18) END) as lp_supply
```
fleupold: As mentioned on the other PR, dividing by 1e18 here causes rounding errors and will make it impossible to get exact results. If this is needed at all it should happen after all summation is done at the end of the query.

cow_amm/cow_amm_lp_and_tvl_4047078
```diff
@@ -0,0 +1,107 @@
...
+, total_lp as (
+select day, lp_supply, sum(lp_supply) over (order by day) as total_lp
+from get_lp_balance
+)
+
+-- lp token total supply without date gaps
+ , lp_total_supply as (
+SELECT
+        date_range.day,
+        COALESCE(v.total_lp, LAG(v.total_lp) OVER (ORDER BY date_range.day)) AS total_lp
```
fleupold: This doesn't work when there is a gap of two days imo (cf. comment on other PR)

cow_amm/cow_amm_lp_and_tvl_4047078
```diff
@@ -0,0 +1,107 @@
...
+    FROM
+        date_range
+    LEFT JOIN
+        total_lp v ON date_range.day = v.day
+)
+-- Compute tvl by multiplying each day's closing price with bet transfers in and out of the pool
```
fleupold:
```suggestion
-- Compute tvl by multiplying each day's closing price with net transfers in and out of the pool
```

cow_amm/cow_amm_lp_and_tvl_4047078
```diff
@@ -0,0 +1,107 @@
...
+, get_tvl as (
+SELECT x.day, SUM(amount * price_close) as total_tvl FROM (
+    SELECT date(evt_block_time) as day, 
+            symbol, 
+            -(value/POW(10,decimals)) as amount
```
fleupold: Also here, I think basic math should be on raw integer amounts to avoid rounding errors you get from using division and floats.

cow_amm/cow_amm_lp_and_tvl_4047078
```diff
@@ -0,0 +1,107 @@
...
+    FROM erc20_ethereum.evt_transfer a LEFT JOIN tokens.erc20 b ON a.contract_address = b.contract_address and blockchain in (select blockchain from cow_amm_pool)
```
fleupold: It doesn't really make sense to support any blockchain here if the table you query from is `erc20_ethereum.evt_transfer` (it won't have any records for gnosis).

We can make a PR later to make these queries multi-chain compatible (let's focus on getting mainnet data for now).

cow_amm/cow_amm_lp_and_tvl_4047078
```diff
@@ -18,31 +38,70 @@ select day, lp_supply, sum(lp_supply) over (order by day) as total_lp
...
+        and (a.contract_address IN (select token_1_address from cow_amm_pool) or 
+        a.contract_address IN (select token_2_address from cow_amm_pool))
+) x join prices.usd_daily y
+    ON blockchain in (select blockchain from cow_amm_pool) and x.symbol = y.symbol and y.day=x.day
+group by 1
 )
 
 , total_tvl as (
-select day,  blockchain, cow_amm_address, tvl as total_tvl
+select day, sum(total_tvl) over (order by day) as total_tvl
```
fleupold: What happens here if a token didn't trade for a day? TVL may have gaps.

olgafetisova: addressed that the same way as for the lp supply

cow_amm/cow_amm_lp_and_tvl_4047078.sql
```diff
@@ -0,0 +1,151 @@
...
+            and lp.day >= (timestamp '{{start}}' - interval '7' day)
+    )
+    where latest = 1
+),
+
+-- Compute tvl by multiplying each day's closing price with net transfers in and out of the pool
+get_tvl as (
+    select
+        x.day,
+        sum(amount * price_close / pow(10, decimals)) as total_tvl
```
fleupold, line 75: This doesn't change anything compared to the logic before (you are still going into floating number before you are summing up all the in and out transfers thus creating imprecision).  

Instead, this needs to become (first summing the whole wei amounts and only then dividing by decimals):
```suggestion
        sum(amount) * price_close / pow(10, decimals) as total_tvl
```


If you want to see an example of the floating imprecision you can run the two python lines which give different output:

```python
(1/6)+(1/6)+(1/6)+(1/6)+(1/6)+(1/6)
(1+1+1+1+1+1)/6
```

cow_amm/cow_amm_lp_and_tvl_4047078.sql
```diff
@@ -0,0 +1,151 @@
...
+    from (
+        select
+            date(evt_block_time) as "day",
+            contract_address,
+            -value as amount
+        from erc20_ethereum.evt_transfer
+        where
+            "from" in (select address from cow_amm_pool)
```
fleupold, line 83: Shouldn't we filter on the token here? Otherwise sending a random token to the USDC/WETH pool would increase its TVL which seems wrong. You can see that this did happen: https://etherscan.io/address/0xb4e16d0168e52d35cacd2c6185b44281ec28c9dc#code

image

fleupold (commented):
Looks good, but I think there are still two subtle issues with the tvl query that should be fixed before merging.

cow_amm/cow_amm_lp_and_tvl_4047078.sql
```diff
@@ -0,0 +1,151 @@
...
+
+-- Finds the CoW AMM pool address given tokens specified in query parameters (regardless of order) 
+cow_amm_pool as (
+    select
+        blockchain,
+        address,
+        token_1_address,
+        token_2_address
+    from query_3959044
+    where ((token_1_address = {{token_a}} and token_2_address = {{token_b}}) or (token_2_address = {{token_a}} and token_1_address = {{token_b}}))
```
fleupold, line 27: What happens here if there is more than one AMM per token pair?

cow_amm/cow_amm_lp_and_tvl_4047078.sql
```diff
@@ -0,0 +1,151 @@
...
+),
+
+-- per day lp token total supply changes of the CoW AMM pool by looking at burn/mint events
+lp_balance_delta as (
+    select
+        date(evt_block_time) as "day",
+        sum(case when "from" = 0x0000000000000000000000000000000000000000 then value else -value end) as lp_supply
+    from erc20_ethereum.evt_transfer
```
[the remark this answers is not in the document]
fleupold: Yes, I asked Olga to focus on mainnet so we have some data and can make sure we are presenting the right thing, but this can be changed to

```
erc20_{{chain}}.evt_transfer
```

and we can make all performance queries chain agnostic (should be a separate PR imo)

cow_amm/cow_amm_lp_and_tvl_4047078.sql
```diff
@@ -0,0 +1,151 @@
...
+    day,
+    tvl,
+    total_lp,
+    lp_token_price,
+    (
+        -- Assess initial investment in lp tokens
+        select 10000 * lp_token_price as investment
+        from final
+        where day = timestamp '{{start}}'
+    ) / lp_token_price as current_value_of_investment
```
fleupold, line 148: Note that the / and * also need to be inverted here

cow_amm/cow_amm_lp_and_tvl_4047078.sql
olgafetisova: That should not be a problem, since we are not selecting a specific pool but rather a token pair, in the next steps we will sum all the tvl and lp supply for all these pools. This is actually the case with the token pair we are testing, so I find it find. Do you see any issues with such approach?

fleupold: Yes, I'm pretty sure the math doesn't check out when adding lp tokens across multiple pools, as lp tokens may have a different price per pool.

Take the following example
- pool a has $100 TVL and 100 LP tokens (1LP token = $1)
- pool b has $100 TVL and 10 LP tokens (1LP token = $10)

If we add this up, we would get $200 TVL for 110 LP tokens, making us believe on LP token is worth $1.8

Now let's say $10 are redeemed. Here it matters in which pool they are redeemed because:
- if redeemed in pool a, it will burn 10 LP tokens (leading to $190TVL and 100 LP tokens)
- if redeemed in pool b, it will burn 1LP token (leading to $190TVL and 109 LP tokens)

In each case we would end up with a different view of how much 1 LP token is worth.

So in summary, we need to fix ourselves to a specific pool (likely the biggest, for now I went with the most recently deployed one in the other query I wrote)

cow_amm/cow_amm_lp_and_tvl_4047078.sql
```diff
@@ -0,0 +1,151 @@
...
+        total_tvl
+    from (
+        -- join full date range with potentially incomplete data. This results in many rows per day (all total tvl on or before that day)
+        -- rank() is then used to order join candidates by recency (rank = 1 is the latest tvl)
+        select
+            date_range.day,
+            total_tvl,
+            rank() over (partition by (date_range.day) order by tvl.day desc) as latest
+        from date_range
+        inner join total_tvl_prep as tvl
```
fleupold, line 117: I created a helper query in #27 which we can use here instead of total_tvl_prep which should simplify this query quite a bit.

cow_amm/cow_amm_lp_and_tvl_4047078.sql
```diff
@@ -0,0 +1,174 @@
...
+            from query_3959044) as p
+            on
+                t.cowamm_address = p.address and t.blockchain = p.blockchain
+                and t.token_1_address = p.token_1_address and t.token_2_address = p.token_2_address
+
+        where
+            ((t.token_1_address = {{token_a}} and t.token_2_address = {{token_b}}) or (t.token_2_address = {{token_a}} and t.token_1_address = {{token_b}}))
+    ) where temp = 1
+    and token_1_balance > 0 and token_2_balance > 0
+),
```
olgafetisova, line 49: Not sure it it makes sense to keep this complex logic to avoid empty test pools. 
The previous query shows the test pools, but it should not affect the result as they are empty.

fleupold: What is this query doing (the comment was not updated to reflect the new logic)? It looks quite bizarre. We are taking some count of addresses of one query as a partitioning key in another one to count the token_1 reserve?

I would simply order by created_at (which is a column that was added to 3959044.

cow_amm/cow_amm_lp_and_tvl_4047078.sql
olgafetisova: But this is not going to work in this query as price and decimals vary per token. So it either has to be done per day per token like this:

```
select day, sum(total_tvl)
from (
select 
        x.day,
        x.contract_address,
        decimals,
        price_close,
        sum(amount) * price_close / pow(10, decimals) as total_tvl
```
Or aggregated.

fleupold (approved):

PR #18, opened 2024-09-09
[CoW AMM Performance] 10k growth for daily rebalancing portfolio

fleupold:
On top of the comparison with the Uniswap reference pool, another interesting reference for CoW AMM performance is the value that a daily rebalancing portfolio (ie a 50:50 ETF that re-allocates once a day) would have over time.

This query achieves this using a recursive SQL statement. It invests 10k on the starting day at the opening price, and then for every consecutive day takes the closing price of the previous day to assess its allocation and perform a rebalancing.

### Test Plan 

Dune run: https://dune.com/queries/4055484

To cross check, I printed the closing prices for each day and redid the calculation manually for a short time period.

### DIFF
--- a/cow_amm/daily_rebalancing_4055484.sql
+++ b/cow_amm/daily_rebalancing_4055484.sql
@@ -1,0 +1,49 @@
+-- Computes the balances and current value of a counterfactual portfolio that invests 10k evenly into two tokens and re-balances once a day to keep a 50:50 exposure
+-- Parameters
+--  {{token_a}} - either token of the desired uni pool
+--  {{token_b}} - other token of the desired uni pool
+--  {{start}} - date as of which the analysis should run
+
+with recursive balances (day, balance0, balance1) as (
+    -- Base case: $10k are invested evenly into token_a and token_b at the opening price of the start day
+    select
+        date(timestamp '{{start}}'),
+        10000 / 2 / p1.price_open, -- Initial balance in token_a
+        10000 / 2 / p2.price_open  -- Initial balance in token_b
+    from prices.usd_daily as p1
+    inner join prices.usd_daily as p2
+        on
+            p1.day = date(timestamp '{{start}}')
+            and p1.day = p2.day
+            and p1.blockchain = 'ethereum'
+            and p2.blockchain = 'ethereum'
+            and p1.contract_address = {{token_a}}
+            and p2.contract_address = {{token_b}}
+
+    union all
+
+    -- Recursive case: Compute the next day's balances according to previous day's closing price and reinvest half into each token
+    select
+        b.day + interval '1' day,
+        (balance0 * p1.price_close + balance1 * p2.price_close) / 2 / p1.price_close, -- Updated balance in token_a
+        (balance0 * p1.price_close + balance1 * p2.price_close) / 2 / p2.price_close  -- Updated balance in token_b
+    from balances as b
+    inner join prices.usd_daily as p1
+        on b.day = p1.day and p1.contract_address = {{token_a}}
+    inner join prices.usd_daily as p2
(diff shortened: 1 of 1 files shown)


harisang:
Looks good. I wonder whether we can work with work with `price_open` and `price_close` of each day so that we don't have to look in the previous day. This would help avoid the last join as we could carry the balances along with the prices in the balances table.

fleupold:
So weirdly enough, prices open of day x + 1 and price close of day x are not guaranteed to be the same. I don't think this join is very expensive, so I wouldn't do it because of that. Do you think it would make the query overall simpler to comprehend?

harisang:
> So weirdly enough, prices open of day x + 1 and price close of day x are not guaranteed to be the same. I don't think this join is very expensive, so I wouldn't do it because of that. Do you think it would make the query overall simpler to comprehend?

Yeah true, i checked myself GNO, that i expected to be really identical but actually there were some small variations. As i am ignorant about "performance improvements" in sql, if you think it's fine, then let's stick to your implementation. Readbility is fine.

harisang:
Btw, do we want to add an end date as well?

fleupold:
Morningstar graphs are always from a start date until the current day, so I left it like that for now, but it would be trivial to add if we wanted to.

harisang (approved):

fhenneke:
There seems be be an issue with maximum recursion length: starting from `2024-08-29 00:00:00` gives the error
```
Error: Recursion depth limit exceeded (10). Use 'max_recursion_depth' session property to modify the limit. [Execution ID: 01J7BNRGBD56MYACCVJZCMP4KS].
```

The code looks clean though.

Using closing and opening prices looks a bit strange. It is supposed to represent rebalancing between closing and opening, right?

fhenneke:
Note that a recursive approach is not required here: for the total value of reserves one gets
```
current_value(day + 1) = current_value(day) * (p1.price_close(day + 1) / p1.price_close(day) + p2.price_close(day + 1) / p2.price_close(day)) / 2
```
with initial condition
```
current_value(start) = 10000 * (p1.price_close(start) / p1.price_open(start) + p2.price_close(start) / p2.price_open(start)) / 2
```
(using opening prices, which does seems a bit unnatural when written like this). Thus, `current_value` can be computed using cumulative products of (sums of) daily price changes
```
(p1.price_close(day + 1) / p1.price_close(day) + p2.price_close(day + 1) / p2.price_close(day)) / 2
```
Balances can be reconstructed from the total value using prices.

fleupold:
> Using closing and opening prices looks a bit strange. It is supposed to represent rebalancing between closing and opening, right?

The goal was to have the first day already capture some price movement (ie. pretending it was deposited at the beginning of the day), but I agree this is a bit confusing.

> Note that a recursive approach is not required here:

You are right, great idea! Will rewrite the query based on that insight.

cow_amm/daily_rebalancing_4055484.sql
```diff
@@ -0,0 +1,58 @@
...
+            previous_p2.day = ds.day - interval '1' day
+            and previous_p2.contract_address = {{token_b}}
+            and previous_p2.blockchain = 'ethereum'
+)
+
+-- For each day multiply initial investment with cumulative product of average price change of the two assets
+select
+    day,
+    -- SQL doesn't support PRODUCT() over (...), but luckily "the sum of logarithms" is equal to "logarithm of the product",
+    exp(sum(ln((p1 + p2) / 2)) over (order by day asc)) * 10000 as current_value
```
fhenneke, line 56: Judging by the two other queries on amm performance, the returned columns might need to be `current_value_of_investment` for consistency.

PR #19, opened 2024-09-09
Add block range, bonding pool and vouch queries

harisang:
This PR migrates queries [3333356](https://dune.com/queries/3333356) and [1541516](https://dune.com/queries/1541516) to the repo. There are also some changes proposed for 1541516, that get rid of the parameter that passes the list of full bonding pools, and proposes instead to use a new query [4056263](https://dune.com/queries/4056263) to hardcode the full bonding pools.

### DIFF
--- a/cowprotocol/accounting/rewards/mainnet/.sqlfluff
+++ b/cowprotocol/accounting/rewards/mainnet/.sqlfluff
@@ -1,4 +1,4 @@
 [sqlfluff:templater:jinja:context]
-start_time='2024-08-20 00:00'
-end_time='2024-08-27 00:00'
+start_time='2024-08-20 00:00:00'
+end_time='2024-08-27 00:00:00'
 
--- a/cowprotocol/accounting/rewards/mainnet/.sqlfluff
+++ b/cowprotocol/accounting/rewards/mainnet/.sqlfluff
@@ -1,0 +1,5 @@
+[sqlfluff:templater:jinja:context]
+start_time='2024-08-20 00:00'
+end_time='2024-08-27 00:00'
+vouch_cte_name=(named_results, valid_vouches)
+
--- a/cowprotocol/accounting/rewards/mainnet/.sqlfluff
+++ b/cowprotocol/accounting/rewards/mainnet/.sqlfluff
@@ -1,5 +1,4 @@
 [sqlfluff:templater:jinja:context]
 start_time='2024-08-20 00:00'
 end_time='2024-08-27 00:00'
-vouch_cte_name=named_results, valid_vouches
 
--- a/cowprotocol/accounting/rewards/mainnet/.sqlfluff
+++ b/cowprotocol/accounting/rewards/mainnet/.sqlfluff
@@ -1,5 +1,5 @@
 [sqlfluff:templater:jinja:context]
 start_time='2024-08-20 00:00'
 end_time='2024-08-27 00:00'
-vouch_cte_name=(named_results, valid_vouches)
+vouch_cte_name=named_results, valid_vouches
 
--- a/cowprotocol/accounting/rewards/mainnet/.sqlfluff
+++ b/cowprotocol/accounting/rewards/mainnet/.sqlfluff
@@ -1,4 +1,5 @@
 [sqlfluff:templater:jinja:context]
 start_time='2024-08-20 00:00:00'
 end_time='2024-08-27 00:00:00'
+vouch_cte_name=valid_vouches,named_results
 
--- a/cowprotocol/accounting/rewards/mainnet/block_number_interval_from_time_interval_query_3333356.sql
+++ b/cowprotocol/accounting/rewards/mainnet/block_number_interval_from_time_interval_query_3333356.sql
@@ -1,0 +1,5 @@
+select
+    min("number") as start_block,
+    max("number") as end_block
+from ethereum.blocks
+where time >= cast('{{start_time}}' as timestamp) and time < cast('{{end_time}}' as timestamp)
--- a/cowprotocol/accounting/rewards/mainnet/block_number_interval_from_time_interval_query_3333356.sql
+++ b/cowprotocol/accounting/rewards/mainnet/block_number_interval_from_time_interval_query_3333356.sql
@@ -1,3 +1,10 @@
+-- Base query to convert accounting period to block range.
+-- Convention throughout the accounting that is used is that we consider all auctions whose
+-- block deadline is in the block range this query returns
+-- Parameters:
+--  {{start_time}} - the start date timestamp for the accounting period  (inclusively)
+--  {{end_time}} - the end date timestamp for the accounting period (exclusively)
+
 select
     min("number") as start_block,
     max("number") as end_block
--- a/cowprotocol/accounting/rewards/mainnet/full_bonding_pools_query_4056263.sql
+++ b/cowprotocol/accounting/rewards/mainnet/full_bonding_pools_query_4056263.sql
@@ -1,0 +1,21 @@
+with
+full_bonding_pools as (
+    select
+        from_hex('0x8353713b6D2F728Ed763a04B886B16aAD2b16eBD') as pool_address,
+        'Gnosis' as pool_name,
+        from_hex('0x6c642cafcbd9d8383250bb25f67ae409147f78b2') as funder
+    union distinct
+    select
+        from_hex('0x5d4020b9261F01B6f8a45db929704b0Ad6F5e9E6') as pool_address,
+        'CoW Services' as pool_name,
+        from_hex('0x423cec87f19f0778f549846e0801ee267a917935') as funder
+    union distinct
+    select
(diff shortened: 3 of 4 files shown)


harisang:
There is still one issues with parsing, as i don't know how to properly pass in sqlfluff this list of table names that we use on Dune to return a different result, depending on the use. Any help would be appreciated

harisang:
Tested them and they seem to work.

cowprotocol/accounting/rewards/mainnet/vouch_registry_query_1541516.sql
```diff
@@ -0,0 +1,104 @@
...
+        bp.pool_name,
+        concat(environment, '-', s.name) as solver_name
+    from valid_vouches as vv
+    inner join cow_protocol_ethereum.solvers as s
+        on address = solver
+    inner join query_4056263 as bp
+        on vv.pool = bp.pool
+)
+
+select * from {{vouch_cte_name}}
```
fhenneke, line 104: Is this variable used in any other query?
Some of our queries seem to have a parameter for selecting different tables within the query. Seems mostly for debugging.

cowprotocol/accounting/rewards/mainnet/vouch_registry_query_1541516.sql
```diff
@@ -0,0 +1,104 @@
+with
```
fhenneke, line 1: Parameters would need documentation.

cowprotocol/accounting/rewards/mainnet/block_number_interval_from_time_interval_query_3333356.sql
```diff
@@ -0,0 +1,5 @@
+select
+    min("number") as start_block,
+    max("number") as end_block
+from ethereum.blocks
+where time >= cast('{{start_time}}' as timestamp) and time < cast('{{end_time}}' as timestamp)
```
fhenneke, line 5: Parameters would need documentation.

cowprotocol/accounting/rewards/mainnet/vouch_registry_query_1541516.sql
```diff
@@ -0,0 +1,104 @@
...
+
+last_block_before_timestamp as (
+    select max(number) from ethereum.blocks
+    where time < cast('{{end_time}}' as timestamp)
+),
```
fhenneke, line 6: This could in principle call out to query 3333356. Then it would be easier to make sure that the convention on which boundary points are included is consistent across queries.

cowprotocol/accounting/rewards/mainnet/vouch_registry_query_1541516.sql
```diff
@@ -0,0 +1,104 @@
...
+
+-- At this point we have excluded all arbitrary vouches (i.e. those not from initial funders of recognized pools)
+-- This ranks (solver, pool, sender) by most recent (vouch or invalidation)
+-- and yields as rank 1, the current "active" status of the triplet.
+vouches_and_invalidations as (
+    select *
+    from vouches
+    union distinct
+    select * from invalidations
```
fhenneke, line 50: styling looks inconsistent

fhenneke (commented):
We could merge this after adding more documentation.

I have not checked the logic of the vouching query.

How would I test that those queries work as intended?

cowprotocol/accounting/rewards/mainnet/vouch_registry_query_1541516.sql
harisang: Yeah good point, it's mainly for getting a different version of the results, but it is not used anywhere else i guess

harisang: I still don't mind having it there tbh

cowprotocol/accounting/rewards/mainnet/vouch_registry_query_1541516.sql
harisang: Yeah thought about it. There is a start_time in query query 3333356, but i guess i could hardcode some very old one to work around it.

cowprotocol/accounting/rewards/mainnet/vouch_registry_query_1541516.sql
harisang: oh thought this would have been fixed automatically. Sooorry

harisang:
> We could merge this after adding more documentation.
> 
> I have not checked the logic of the vouching query.
> 
> How would I test that those queries work as intended?

Good question. Ok so 2 of the 3 queries are very straightforward and easy to test. The vouching query is essentially the one we currently use, but i believe it actually might not be 100% correct, as it might not take into account cases where a solver left a bonding pool and then rejoined during the accounting period we are interested in.

I will open an issue for this one, as i don't have the capacity to fix this at the moment.

harisang:
Opened issue #21  in order to revisit the logic of the vouching query.

cowprotocol/accounting/rewards/mainnet/vouch_registry_query_1541516.sql
fhenneke: I there is a real use case we can keep it and document it.
Otherwise it just makes the query more difficult to understand.

cowprotocol/accounting/rewards/mainnet/full_bonding_pools_query_4056263.sql
```diff
@@ -0,0 +1,23 @@
...
+with
+full_bonding_pools as (
+    select
+        from_hex('0x8353713b6D2F728Ed763a04B886B16aAD2b16eBD') as pool,
+        'Gnosis' as pool_name,
+        from_hex('0x6c642cafcbd9d8383250bb25f67ae409147f78b2') as initial_funder
+    union distinct
+    select
+        from_hex('0x5d4020b9261F01B6f8a45db929704b0Ad6F5e9E6') as pool,
+        'CoW Services' as pool_name,
```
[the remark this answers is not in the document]
fleupold: yes, this should be call CoW DAO I believe

cowprotocol/accounting/rewards/mainnet/vouch_registry_query_1541516.sql
harisang: Removed it.

harisang:
@fhenneke I did some review of the query and it makes sense. I am not very comfortable with the `rank() over` commands used in 58-61 and 71-74 but overall it looks ok.

Let me know if you want to do another pass

cowprotocol/accounting/rewards/mainnet/.sqlfluff
```diff
@@ -0,0 +1,5 @@
+[sqlfluff:templater:jinja:context]
+start_time='2024-08-20 00:00'
+end_time='2024-08-27 00:00'
+vouch_cte_name=named_results, valid_vouches
```
fhenneke, line 4:
```suggestion
```
as it is not used in the query anymore

cowprotocol/accounting/rewards/mainnet/vouch_registry_query_1541516.sql
fhenneke, line 1: We can add this as is to stay compatible with its current function.

An alternative, which I prefer, is to simplify it to just get the latest pool and reward address, irrespective of any invalidations later on. Solvers changing bonding pools or getting unvouched before running the script have been a pain.

cowprotocol/accounting/rewards/mainnet/.sqlfluff
harisang: ah good point

cowprotocol/accounting/rewards/mainnet/vouch_registry_query_1541516.sql
harisang: Just to clarify, are you saying we should simply fetch the latest vouching event of every solver, and then don't worry about the rest, as in case the solver is later unvouched, we will have in any case blacklisted as well and then that solver would have no effect in the competition?

fhenneke: As far as uses in solver rewards, I think we should only care about latest reward addresses (and bonding pools). If a solver gets blacklisted and removed, solver rewards should still be rewarded as if they had not been removed. Blacklisting should take care of barring solvers from gaining additional rewards if they are not eligible.

Other places might use that query differently. But for our main use case, this simplification should be useful.

harisang: yeah makes sense, will proceed with the simplification

harisang: Actually, i think i changed my mind on this again. This query also shows in the main dashboard, and i think it is nice to be able to see all "active" solvers, with their names etc. So if you don't mind, i would keep it as is and would actually even revert to the previous state where there are two views of the results, so that we can use the other view in the main dashboard.

PR #23, opened 2024-09-10
[CoW AMM] Metric for daily invariant growth (surplus) per $100

fleupold:
As an additional metric next to "growth of 10k" we want to compare the invariance growth of CoW AMMs with the ones from Uniswap.

For CoW AMM this is equal to surplus, for Uniswap it's equal to the swap fee amount (volume * 0,3%). Both are normalized over TVL in order to have a reasonable comparison.

This query adds the comparison for CoW AMM.

### Test Plan

Comparing https://balancer.fi/pools/ethereum/cow/0xf08d4dea369c456d26a3168ff0024b904f2d8b91 with https://dune.com/queries/4059213 for a rough sanity check (values cannot match exactly because balancer uses a different price feed)


cow_amm/profitability/invariant_growth/cow_amm_4059213.sql
```diff
@@ -0,0 +1,92 @@
...
+        token_1_address,
+        token_2_address
+    from query_3959044
+    where ((token_1_address = {{token_a}} and token_2_address = {{token_b}}) or (token_2_address = {{token_a}} and token_1_address = {{token_b}}))
+    order by 1 desc
+    limit 1
+),
+
+-- Computes the token balance changes of the relevant token per transaction
+reserves_delta as (
```
fhenneke, line 21: Most of the query can be substitutes by calling the base query.

The base query also tries to take care of
* liquidity injections
* deposits and withdrawals

If it is actually this easy, the simplifications should be added to the base query as well.

fleupold: Which base query do you mean exactly? I created #27 (but am not aware of an existing one)

fhenneke: I meant [query 3968186](https://dune.com/queries/3968186), which is currently used for dashboards. It is not currently part of this repo.

PR #24, opened 2024-09-10
[Draft, still broken] Fix ordering in rebalancing query

fhenneke:
Ordering of the full table seems to not play well with the ordering in the cumulative sum.

Her I just changed the ordering from `desc` to `asc`.


fhenneke:
The query seems fine after all (though the initial guess creates some issue).

PR #25, opened 2024-09-10
[CoW AMM] Fix uni 10k growth query

fleupold:
The original query was multiplying the initial assessment with the lp_token_price. This is wrong. If we want to compute the lp token amount we can get for 10k, we need to divide.

Similarly, in order to get the current value given the initial amount of lp tokens we need to multiply.


fhenneke (approved):
LGTM

PR #26, opened 2024-09-10
Make daily rebelancing query start at 10k

fleupold:
Otherwise the investment of 10k graph looks off:

Before
image

After
image


fhenneke (approved):
LGTM

PR #30, opened 2024-09-12
Fix lookback range for uni lp token supply

fleupold:
Assuming an lp token event every 7 days was too optimistic for pools like MKR/ETH:

image

This PR changes the performance optimisation by adding another fast subquery to identify the last day we need to consider for the join. This should ensure that the query still runs fast for very active AMMs such as ETH/USDC

### Test Plan
- MKR/ETH now shows data
- USDC/ETH still runs in ~1 minute


harisang:
Didn't test it but makes sense

PR #31, opened 2024-09-16
Refactor slippage query

fhenneke:
This PR refactors the slippage query. It separates different parts of the slippage computation, simplifies debugging unusual slippage, and prepares for moving some of the slippage computation offline.

### Query layout

The general layout of the query is as follows (dashed lines and rectangles correspond to queries which do not yet exist):

Slippage drawio(5)

* There are three mostly independent queries on raw token imbalances, fees, and prices. Those correspond to tables we plan to have available offline.
  * Raw imbalances are computed from all balance changes. They are observable on chain and can be computed on Dune for all chains we support.
  * Fees are computed from offchain information we upload. Fees are classified into `'protocol_fee'` and `'network_fee'`. Protocol fees could in future PRs be separated into protocol fees and partner fees, and, combined with other prices, could give totals for protocol and parner fees.
  * Prices are hourly averages. If Dune prices are available, those prices are used. If Dune prices are not available, prices are reconstructed from trades involving that token with another token which does have a Dune price. Only if that is not possible, the price is null.
* The results of those three queries are combined into a raw slippage query. It has raw imbalances and explicit entries for fees, as well as aggregated numbers per transaction. This table is useful for detecting wrong prices and wrong imbalances, as well as having an unfiltered view of slippage per transaction.
* The result of the previous query is filtered using excluded batches and aggregated pe solver. This query can be used to detect cases of unusual slippage and also gives totals used for payments. Solvers can use that query and filter on their solver address to detect large slippage settlements per solver.
* Since the format of the result is not exactly what the solver-rewards script expects, there is another query, an updated old slippage query, which adds some meta information and renames columns. The naming convention in solver rewards should be changed and then this query can be removed.

The changes to the main slippage query can be observed [here](https://dune.com/queries/4074138).

Except for the empty fee table, all of the queries should work on other chains as well. Missing fee data currently means that network and protocol fees are treated as slippage, which is fine for network fees but incorrect for protocol fees.

### Differences to old slippage query

This is mostly but not quite a replacement of the old slippage query. To exceptions:
* The old query computed prices from exchange rates of the auction. It did so, however, in a way which broke on a lot of cases due to missing data on decimals. The new query works with prices per atom which do not require decimals and can thus recover more token prices from trades.
* The old query for slippage per transaction (but not the one in the repo at the moment, due to a bug) had a parameter for filtering on solver address. This is removed for simplicity. Instead one can filter the final table in the web interface.

### ~Fix of balances changes query~

~Working on slippage revealed an issue with the balance changes query: unlike on mainnet, it is not required on Gnosis to have special treatment for sDAI. All balance changes are accompanied by transfer events. Looking at deposit and withdraw events is not required.~

~The change is also independently addresses in #41.~

(Updated with modified design.)


cowprotocol/accounting/fees/classified_fees_4058574.sql
```diff
@@ -0,0 +1,60 @@
...
+        atoms_sold,
+        atoms_bought,
+        sell_token_address,
+        cast(cast(data.protocol_fee as varchar) as int256) as protocol_fee, -- noqa: RF01
+        cast(cast(data.surplus_fee as varchar) as int256) as surplus_fee, -- noqa: RF01
+        from_hex(data.protocol_fee_token) as protocol_fee_token_address -- noqa: RF01
+    from cow_protocol_{{blockchain}}.trades as t
+    inner join cowswap.raw_order_rewards as ror -- this table only exists for ethereum trades
+        on t.order_uid = from_hex(ror.order_uid) and t.tx_hash = from_hex(ror.tx_hash)
+    where block_time >= cast('{{start_time}}' as timestamp) and block_time < cast('{{end_time}}' as timestamp)
```
harisang, line 32: should we invoke the block range query here so that we are consistent throughout?

cowprotocol/accounting/fees/classified_fees_4058574.sql
```diff
@@ -0,0 +1,60 @@
...
+--
+-- The columns of the result are
+-- - block_time: time of settlement transaction
+-- - tx_hash: settlement transaction hash
+-- - token_address: address of token with a balance change. contract address for erc20 tokens,
+--   0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee for native token
+-- - amount: value of fee in atoms of the token
+-- - transfer_type: 'protocol_fee' for the total protocol fee (including partner fee), 'network_fee' for network fees
+
+with fee_data as (
```
harisang, line 18: should we call this maybe raw_fee_data as, example, the "surplus_fee" value is included here?

fhenneke: Sure

cowprotocol/accounting/fees/classified_fees_4058574.sql
fhenneke: Good question.

Being consistent would be nice. It would also mean that the imbalances query would need to be updated/could not be used.

I would prefer to move towards time based accounting periods everywhere (i.e. using block_number instead of block_deadline (or rather coalesce(block_number, block_deadline) to handle penalties).

Since slippage is (or rather should in principle be) independent of rewards, having inconsistent time ranges is not such a big deal. We have a similar time based accounting for quote rewards. It is arguably more annoying for quote rewards as part of rewards.

cowprotocol/accounting/slippage/slippage_breakdown_4059683.sql
```diff
@@ -0,0 +1,73 @@
+-- This query a breakdown of slippage on CoW Protocol
```
harisang, line 1: fix comment

cowprotocol/accounting/fees/classified_fees_4058574.sql
```diff
@@ -0,0 +1,60 @@
...
+        on t.order_uid = from_hex(ror.order_uid) and t.tx_hash = from_hex(ror.tx_hash)
+    where block_time >= cast('{{start_time}}' as timestamp) and block_time < cast('{{end_time}}' as timestamp)
+),
+
+protocol_fee_balance_changes as (
+    select
+        block_time,
+        tx_hash,
+        protocol_fee_token_address as token_address,
+        protocol_fee as amount,
```
harisang, line 40: Does the order of columns appearing affects the union at lines 58-60? In any case it would be nice to make it consistent with the table `network_fee_balance_changes` by pushing the `amount` here at the very end.

cowprotocol/accounting/fees/classified_fees_4058574.sql
```diff
@@ -0,0 +1,60 @@
...
+--  {{end_time}} - the timestamp for which the analysis should end (exclusively)
+--  {{blockchain}} - network to run the analysis on
+--
+-- The columns of the result are
+-- - block_time: time of settlement transaction
+-- - tx_hash: settlement transaction hash
+-- - token_address: address of token with a balance change. contract address for erc20 tokens,
+--   0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee for native token
+-- - amount: value of fee in atoms of the token
+-- - transfer_type: 'protocol_fee' for the total protocol fee (including partner fee), 'network_fee' for network fees
```
harisang, line 16: it does look a bit strange that this column is labeled `transfer_type` although the query is meant to just compute fees. Should we maybe name it just `type` and then rename it in the context of slippage accounting?

cowprotocol/accounting/fees/classified_fees_4058574.sql
harisang: Ok will continue reviewing but then i would actually suggest we move everything to block numbers and only have the top-level query use dates

cowprotocol/accounting/slippage/slippage_breakdown_4059683.sql
```diff
@@ -0,0 +1,73 @@
...
+        date_trunc('hour', block_time) as hour --noqa: RF04
+    from "query_4021644(blockchain='{{blockchain}}',start_time='{{start_time}}',end_time='{{end_time}}')"
+),
+
+fee_balance_changes as (
+    select
+        block_time,
+        tx_hash,
+        token_address,
+        transfer_type,
```
harisang, line 38: Same as comment above that i would expect the order of columns to be the same as the one used in `raw_token_imbalances`

cowprotocol/accounting/slippage/slippage_breakdown_4059683.sql
```diff
@@ -0,0 +1,73 @@
...
+-- - token_address: address of token with slippage. contract address for erc20 tokens,
+--   0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee for native token
+-- - amount: signed value of slippage in atoms of the token; fees are represented as negative
+--   integers since they will be removed from imbalances
+-- - transfer_type: 'raw_imbalance' for imbalance observable on chain, 'protocol_fee' for the total
+--   protocol fee (including partner fee), 'network_fee' for network fees
+-- - price: USD price of one unit (i.e. pow(10, decimals) atoms) of a token
+-- - price_atom: USD price of one atom (i.e. 1. / pow(10, decimals) units) of a token
+-- - slippage_usd: USD value of slippage
+-- - slippage_native_atom: value of slippage in atoms of native token
```
harisang, line 20: I find the name `slippage_native_atom` a bit confusing. I would actually go for the very self-explanatory
`slippage_in_native_token_atoms`

And then i would also use `price_in_usd`

cowprotocol/accounting/slippage/slippage_breakdown_4059683.sql
```diff
@@ -0,0 +1,73 @@
...
+-- The columns of the result are
+-- - block_time: time of settlement transaction
+-- - tx_hash: settlement transaction hash
+-- - token_address: address of token with slippage. contract address for erc20 tokens,
+--   0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee for native token
+-- - amount: signed value of slippage in atoms of the token; fees are represented as negative
+--   integers since they will be removed from imbalances
+-- - transfer_type: 'raw_imbalance' for imbalance observable on chain, 'protocol_fee' for the total
+--   protocol fee (including partner fee), 'network_fee' for network fees
+-- - price: USD price of one unit (i.e. pow(10, decimals) atoms) of a token
```
harisang, line 17: why not use `price_unit` so that it is clear what's the difference between this and the `price_atom`

cowprotocol/accounting/slippage/slippage_per_solver_4070065.sql
```diff
@@ -0,0 +1,20 @@
+-- This query return slippage per transaction
```
harisang, line 1:
```suggestion
-- This query returns slippage per transaction, evaluated in both usd and the native token of the chain.
```

cowprotocol/accounting/slippage/slippage_per_solver_4070065.sql
```diff
@@ -0,0 +1,20 @@
...
+--
+-- Parameters:
+--  {{start_time}} - the timestamp for which the analysis should start (inclusively)
+--  {{end_time}} - the timestamp for which the analysis should end (exclusively)
+--  {{blockchain}} - network to run the analysis on
+--
+-- The columns of the result are
```
harisang, line 8: The next few lines also need fixing

cowprotocol/accounting/slippage/slippage_per_transaction_4070059.sql
```diff
@@ -0,0 +1,33 @@
+-- This query return slippage per transaction
```
harisang, line 1:
```suggestion
-- This query returns slippage per transaction, evaluated in both usd and the native token of the chain.
```

cowprotocol/accounting/slippage/slippage_per_transaction_4070059.sql
```diff
@@ -0,0 +1,33 @@
...
+--
+-- The columns of the result are
+-- - block_time: time of settlement transaction
+-- - tx_hash: settlement transaction hash
+-- - solver_address: address of the solver executing the settlement
+-- - slippage_usd: USD value of slippage
+-- - slippage_native: value of slippage in native token
+--
+-- Results of the query are filtered to not include batches from excluded_batches.
+-- Batches are also excluded if there is a non-zero imbalance and no value (in native atomes).
```
harisang, line 16:
```suggestion
-- Batches are also excluded if there is a non-zero imbalance and no value (in native atoms).
```

cowprotocol/accounting/slippage/slippage_per_transaction_4070059.sql
```diff
@@ -0,0 +1,33 @@
...
+
+with excluded_batches as (
```
harisang, line 18: I think this should be moved to the `slippage_per_solver` query.

cowprotocol/accounting/slippage/slippage_per_transaction_4070059.sql
```diff
@@ -0,0 +1,33 @@
...
+    select tx_hash from query_3490353
+)
+
+select
+    s.block_time,
+    s.tx_hash,
+    solver_address,
```
harisang, line 25: i guess solver_address could be skipped here as it is not really relevant to the main functionality of the query.

cowprotocol/accounting/slippage/slippage_prices_4064601.sql
```diff
@@ -0,0 +1,131 @@
...
+with token_times as (
+    select
+        token_address,
+        date_trunc('hour', block_time) as hour --noqa: RF04
+    from "query_4021257(blockchain='{{blockchain}}',start_time='{{start_time}}',end_time='{{end_time}}')"
+    where token_address != 0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee
+    group by 1, 2
+),
+
+precise_prices as (
```
harisang, line 27: maybe let's add some comments to describe what each of the following tables computes

cowprotocol/accounting/slippage/slippage_per_transaction_4070059.sql
fhenneke: One goal was to have a query per transaction which is exactly what we use for total slippage to base an unusual slippage query on.

If you think that is not a requirement and having a discrepancy there if fine, I can move excluded batches somewhere else.

Maybe the better distinction is to merge the three queries slippage breakdown, slippage per transaction, slippage per solver into
- raw slippage (including breakdown and aggregation per transaction)
- slippage (with filtering, per transaction and per solver)

The first query would then be for debugging and could also leave out solver names, the second would be compatible with actual payment.

cowprotocol/accounting/slippage/slippage_per_transaction_4070059.sql
fhenneke: It was useful for me to have a query with results per transaction and where I could filter on solver address. A cleaner way might be what I proposed in the response above.

cowprotocol/accounting/slippage/slippage_per_transaction_4070059.sql
```diff
@@ -0,0 +1,33 @@
...
+    sum(slippage_usd) as slippage_usd,
+    sum(slippage_native_atom) as slippage_native_atom
+from "query_4059683(blockchain='{{blockchain}}',start_time='{{start_time}}',end_time='{{end_time}}')" as s
+inner join cow_protocol_{{blockchain}}.batches as b
+    on s.tx_hash = b.tx_hash
+where s.tx_hash not in (select tx_hash from excluded_batches)
+group by 1, 2, 3
+having bool_and(slippage_native_atom is not null or slippage_atoms = 0)
```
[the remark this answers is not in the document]
fhenneke: I also did not know this before I saw it being used in the old slippage query.
`bool_and` is an aggregating expression, similar to `sum`. it checks that all of the expressions (iterating over rows within a group, same as with `sum`), i.e. the logical `and` of all the entries.

Here it is used to check that for each entry in the slippage breakdown `slippage_wei is not null`, i.e., we have a value for slippage in wei due to a price being available, or `slippage_atoms = 0`, i.e., there is no slippage so we also do not need a price. This excludes all settlements where a price is missing for a token for which we need a price.

cowprotocol/accounting/fees/classified_fees_4058574.sql
fhenneke: I think this is not possible, unfortunately. Query parameters in queries you call from within another query cannot be computed values. One can easily forward time ranges as they are not computed but parameters of the calling query. Passing block ranges would require computing them from dates.

Not sure if there is another way.

cowprotocol/accounting/fees/classified_fees_4058574.sql
fhenneke: I changed the ordering to be consistent.

fhenneke:
I joined the new and old slippage query to check differences, https://dune.com/queries/4083184.

In cases where both queries report a value, they are almost identical (~1e-10).
The new query returns a slippage value for more auctions. Looking at the largest and smallest slippage values indicates that the new query is more accurate.

- positive slippage:
  - https://etherscan.io/tx/0xb003b9cf3d448bfbbb6161519504cd9f761fde1cef0ca4e544ef7daa44d5a944
$90 of slippage, price missing for 0xd39de8fcecb7f2e0c3deb198d8e33663afa0e977, which can be reconstructed from the auction
  - https://etherscan.io/tx/0xdce056e27585d11304ed70282b677bf59f7c7648ee4c951bbcb7de70709b8d5b
$30 of slippage, price missing for 0xb6f4d11d6c274f8f0a55d101b65223a45a3d1128, which can be reconstructed from the auction

- negative slippage:
  - https://etherscan.io/tx/0x4d38fe407beb4cef9651913f73adab65529df7daf0d968256e48fe453667d262
-$30 of slippage, price missing for 0x88ce174c655b6d11210a069b2c106632dabdb068, which can be reconstructed from the auction
amounts cancel almost exactly, but not quite (due to rounding in protocol fees?)
  - https://etherscan.io/tx/0x7ca4094e1ae5a7d68c571ceecb247efcd66f1953cf7c81c533465dd847410f80
-$30 of slippage, price missing for 0x777be1c6075c20184c4fd76344b7b0b7c858fe6b, which can be reconstructed from the auction. the price of $2.62 closely corresponded to the coingecko price $2.63 at the time of the auction

cowprotocol/accounting/fees/classified_fees_4058574.sql
```diff
@@ -0,0 +1,60 @@
+-- This query computes fees
+--
+-- It currently uses the raw_order_rewards table which is only available on ethereum. On other chains it returns an empty table.
+--
+-- Parameters:
+--  {{start_time}} - the timestamp for which the analysis should start (inclusively)
+--  {{end_time}} - the timestamp for which the analysis should end (exclusively)
+--  {{blockchain}} - network to run the analysis on
+--
+-- The columns of the result are
```
harisang, line 10: since the result is really on a per order basis, wouldn't it make sense to keep the uid there as well.  I know it's irrelevant for the actual slippage calculations later on but if we claim this query computes fees on a per trade basis, it is a bit strange not to include the uid

cowprotocol/accounting/fees/classified_fees_4058574.sql
```diff
@@ -0,0 +1,60 @@
...
+        cast(cast(data.protocol_fee as varchar) as int256) as protocol_fee, -- noqa: RF01
+        cast(cast(data.surplus_fee as varchar) as int256) as surplus_fee, -- noqa: RF01
+        from_hex(data.protocol_fee_token) as protocol_fee_token_address -- noqa: RF01
+    from cow_protocol_{{blockchain}}.trades as t
+    inner join cowswap.raw_order_rewards as ror -- this table only exists for ethereum trades
+        on t.order_uid = from_hex(ror.order_uid) and t.tx_hash = from_hex(ror.tx_hash)
+    where block_time >= cast('{{start_time}}' as timestamp) and block_time < cast('{{end_time}}' as timestamp)
+),
+
+protocol_fee_balance_changes as (
```
harisang, line 35: The table name, just looking at this query, is a bit strange. Same for the next one

cowprotocol/accounting/slippage/raw_slippage_4059683.sql
```diff
@@ -0,0 +1,96 @@
...
+    select * from raw_token_imbalances
+    union all
+    select * from fee_balance_changes
+),
+
+prices as (
+    select * from "query_4064601(blockchain='{{blockchain}}',start_time='{{start_time}}',end_time='{{end_time}}')"
+),
+
+raw_slippage_breakdown as (
```
harisang, line 62: what do we mean by "raw" here?

fhenneke: I want to differentiate between slippage values without much processing ("raw_slippage", for debugging) and slippage values after processing (just "slippage", what is really used later on).

Maybe "unfiltered" would be better. Or just slippage_breakdown and slippage_per_transaction and name the query "slippage_breakdown"?

Do you have a different naming suggestion?

harisang:
Looks good!

> Working on slippage revealed an issue with the balance changes query: unlike on mainnet, it is not required on Gnosis to have special treatment for sDAI. All balance changes are accompanied by transfer events. Looking at deposit and withdraw events is not required. The change is also independently addresses in https://github.com/cowprotocol/dune-queries/issues/41.


Shouldn't with then remove that change from here?

fhenneke:
> Shouldn't` with then remove that change from here?
I will remove them now. They were useful for testing on Gnosis since otherwise wrong slippage would have dominated everything.

harisang (approved):

PR #39, opened 2024-09-17
Sync dashboard query with what is on Dune

harisang:
Some last-minute changes took place today so this PR addresses this.


fleupold:
I think this is a duplicate of #40 (which has a better description of what this change actually is) or am I missing something?

harisang:
> I think this is a duplicate of #40 (which has a better description of what this change actually is) or am I missing something?

Yeah indeed, so closing this one

PR #42, opened 2024-09-17
[CoW AMM] Fix TVL calculation for 10k growth analysis

fleupold:
The TVL was taken directly from [this helper query](https://dune.com/queries/4059700). However its $ value represents the TVL at the time of the last trade (not at the end of each day). This led to CoW AMM not tracking the rebalancing and Uniswap portfolio as closely as it should.

This PR instead uses the raw token balances and joins them with up to date closing prices for each day to get a more correct TVL.

### Test Plan

New USDC/WETH comparison:
image


acanidio-econ:
looks good to me! But I'll let @harisang or @fhenneke also look at it

harisang (approved):
Looks good!

PR #43, opened 2024-09-18
Add CoW AMM surplus bounty query

harisang:
This PR adds (and slightly updates) the query ([4031724](https://dune.com/queries/4031724)) we will use for distributing the service fee collected to solvers that generate surplus for CoW AMMs via CoWs.


harisang:
Tested the query and it seems to be working

fhenneke (approved):
I only added minor comments.
I have not tested the result.

PR #46, opened 2024-09-20
Support CoW AMM stats on all chains

fleupold:
This PR makes it so that statistics for CoW AMMs are chain agnostic. Unfortunately I wasn't able to get the Uniswap graph working our of the box, as it uses different event tables on different chains (and dunes variable syntax isn't flexible enough to choose a target table based on a condition).

Will ask them on tg, but for now we will have to live with CoW AMM performance vs. rebalancing portfolio.

### Test Plan

- Arbitrum: https://dune.com/cowprotocol/cow-amm-micro-v2?token_a=0x2c650dab03a59332e2e0c0c4a7f726913e5028c1&token_b=0x82af49447d8a07e3bd95bd0d56f35241523fbab1&blockchain=arbitrum&blockchain_t62d7a=arbitrum&token_a_t4bf20=0x2c650dab03a59332e2e0c0c4a7f726913e5028c1&token_b_t51de7=0x82af49447d8a07e3bd95bd0d56f35241523fbab1&start_d87a1a=2024-09-17+00%3A00%3A00
- Gnosis: https://dune.com/cowprotocol/cow-amm-micro-v2?token_a=0x6a023ccd1ff6f2045c3309768ead9e68f978f6e1&token_b=0x9c58bacc331c9aa871afd802db6379a98e80cedb&blockchain=gnosis


fhenneke:
> Unfortunately I wasn't able to get the Uniswap graph working our of the box, as it uses different even tables on different chains (and dunes variable syntax isn't flexible enough to choose a target table based on a condition).

Would the approach in https://github.com/cowprotocol/dune-queries/blob/d7f5e6b6a0c4931290c517d5fb939b2db36dffba/cowprotocol/accounting/token_imbalances/balance_changes_4021257.sql#L241 help here?

It is a bit ugly but one can create different tables for different chains and then select the correct one based on a parameter.

fleupold:
This could work. I'm also considering reconstructing the events based on raw logs (which is available in a chain uniform way). However, I think it's ok to have a separate PR for this.

harisang (approved):
Looks good. Would only update the description at the top of 3959044 and also fix the typo in the description of 4059700

fleupold:
> and also fix the typo in the description of 4059700

It's usually easier to comment directly on the line that has the type so that I don't have to guess which one you mean (although it's a good test for my orthographic reading skills). You can even suggest the fix there :-)

PR #47, opened 2024-09-20
[CoW AMM] 10k growth for HODL strategy

fleupold:
Not a super meaningful but fun and simple metric to compute, this query plots the 10k growth graph of "not LPing" (ie. holding the two assets). The difference is usually referred to as "impermanent loss".

### Test Plan

https://dune.com/queries/4059180/6834667

image

### DIFF
--- a/cowamm/profitability/10k_growth/hodl_4086902.sql
+++ b/cowamm/profitability/10k_growth/hodl_4086902.sql
@@ -30,7 +30,8 @@ starting_balance as (
 select
     ds.day,
     token_a_start * p1.price_close + token_b_start * p2.price_close as current_value_of_investment
-from starting_balance, date_series as ds
+from starting_balance
+cross join date_series as ds
 inner join prices.usd_daily as p1
     on
         ds.day = p1.day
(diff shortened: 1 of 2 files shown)


cowamm/profitability/10k_growth/hodl_4086902.sql
```diff
@@ -0,0 +1,42 @@
...
+            p1.day = p2.day
+            and p1.day = date(timestamp '{{start}}')
+            and p1.contract_address = {{token_a}}
+            and p2.contract_address = {{token_b}}
+)
+
+select
+    ds.day,
+    token_a_start * p1.price_close + token_b_start * p2.price_close as current_value_of_investment
+from starting_balance, date_series as ds
```
harisang, line 33: I admit i had never seen this comma notation, and i would probably prefer a more explicit join but i tested it and it works so i am fine with it

harisang (approved):

cowamm/profitability/10k_growth/hodl_4086902.sql
fleupold: replaced with cross join (which is semantically equivalent)

PR #48, opened 2024-09-23
Cowamm/univ2 multichain

fleupold:
The curated events tables don't follow a consistent format making it hard to abstract away the chain for different networks. Luckily the logic for indexing base events is very simple and can be done using the raw `logs` table (which does follow consistent naming.

### Changes
- [x] replace PairCreated, Sync and Swap tables with parsing raw ethereum logs
- [x] use new cte's for invariant growth and 10k growth queries

### Test Plan

We now have reference pool data on [Gnosis Chain](https://dune.com/cowprotocol/cow-amm-micro-v2?token_a=0x6a023ccd1ff6f2045c3309768ead9e68f978f6e1&token_b=0x9c58bacc331c9aa871afd802db6379a98e80cedb&blockchain=gnosis)

image


cowamm/profitability/10k_growth/uni_v2_4047194.sql
```diff
@@ -17,15 +17,43 @@ with date_range as (
 ),
 
 -- Finds the uniswap v2 pool address given tokens specified in query parameters (regardless of order)
-pool as (
+pools as (
```
harisang, line 20: Is it possible that a pool is created, then closed, and then created again? If so, i guess this table would pick on all of these created pools. Not sure if it can ever happen but in that case i guess we should pick the latest creation log.

harisang (approved):
Looks good! Asked one clarifying question but other than that i don't see any problems.

There is some code duplication so we could have the sync and swap events computed in some other subquery but that can be done in another PR, if we think it makes sense.

fleupold:
Combined the two queries

PR #49, opened 2024-09-23
[EASY] Change slippage wrapper for consistency with old slippage query

fhenneke:
This PR makes the slippage query wrapper consistent with the old slippage query.

- The `blockchain` parameter is set to `ethereum`. With this change, solver rewards does not have to add an additional parameter right now when querying slippage.
- The return type is changed to floating point numbers. This removes the need to have an additional conversion from within solver rewards.

This wrapper will be removed once we have to time to adapt the solver rewards script to the format of the new slippage query (or we remove that query and use a local database).

### DIFF
--- a/cowprotocol/accounting/slippage/slippage_query_3427730.sql
+++ b/cowprotocol/accounting/slippage/slippage_query_3427730.sql
@@ -2,7 +2,7 @@ select
     solver_address,
     concat(environment, '-', name) as solver_name,
     slippage_usd as usd_value,
-    slippage_wei as eth_slippage_wei,
+    1.0 * slippage_wei as eth_slippage_wei,
     concat(
         '<a href="https://dune.com/queries/4070065',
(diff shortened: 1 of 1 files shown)


harisang:
@fhenneke I don't see any change related to this

> The return type is changed to floating point numbers. This removes the need to have an additional conversion from within solver rewards.

harisang (changes requested):
Probably forgot to include the edit on the slippage query

fhenneke:
> Probably forgot to include the edit on the slippage query

True. I pushed that fix as well now.

harisang (approved):
Approving but i guess this fix should be either on the 4070065 query side or in the payouts script itself

fhenneke:
I think the slippage query itself should use integers.

At some point we need to change the solver rewards script to handle those values correctly. Until then, this wrapper provides compatibility with the current version of solver rewards.

PR #50, opened 2024-09-24
[CoW AMM] query for total TVL

fleupold:
Query to fetch total TVL for all CoW AMMs on a single chain. Used in the main KPI dashboard to plot current TVL, TVL growth over 7 days and TVL over time.

It's similar to `tvl_by_tx_4059700` but I didn't find a good way to combine the two efficiently.

### Test Plan
https://dune.com/cowprotocol/cow-amms-v2

image


harisang:
You also added the move of an irrelevant query (the one that computes CoW AMM surplus from CoWs with users) to this PR and i would probably not include it here as it has nothing to do with the PR itself.

cowamm/kpi/tvl_by_chain_4096107.sql
```diff
@@ -0,0 +1,85 @@
...
+-- Parameters
+--  {{blockchain}} - the target chain
+
+with pools as (
+    select
+        created_at,
+        token_1_address,
+        token_2_address,
+        address
+    from query_3959044
```
harisang, line 11: Shouldn't we filter based on blcokchain here? There is a blockchain column in the table returned from 3959044 so that is straightforward to do

cowamm/kpi/tvl_by_chain_4096107.sql
```diff
@@ -0,0 +1,85 @@
...
+),
+
+date_range as (
```
harisang, line 14: Should we distinguish between mainnet/gnosis and Arbitrum, which launched later?

fleupold: I think it's fine to start the TVL analysis at the earliest possible day (it will just be 0 for chains that haven't launched yet)

harisang (approved):

PR #51, opened 2024-09-26
Update CoW AMM service fee bounty query to include all chains

harisang:
This PR updates query 4031724 so that it aggregates surplus over all chains, which is generated from CoWs between CoW AMMs and user orders. The [query](https://dune.com/queries/4031724) on Dune is already updated as we needed to use it to payout this week's bounty. Still a proper review would be ideal.

### DIFF
--- a/cowamm/cow_amm_cow_with_users_surplus_query_4031724.sql
+++ b/cowamm/cow_amm_cow_with_users_surplus_query_4031724.sql
@@ -1,6 +1,6 @@
 -- This query computes how much surplus has been provided to CoW AMMs, when trading with other user orders
 -- as part of a CoW. For that, a CoW detector query is used (4025739(). Finally, the query computes the 
--- distribution of an amount {{budget}} of COW tokens to solvers, proportionally to the surplus generated 
+-- distribution of an amount {{cow_budget}} of COW tokens to solvers, proportionally to the surplus generated 
 -- via CoWs and pushed to CoW AMMs.
 -- Parameters:
 --  {{start_time}} - the start date timestamp for the accounting period  (inclusively)
--- a/cowamm/cow_amm_cow_with_users_surplus_query_4031724.sql
+++ b/cowamm/cow_amm_cow_with_users_surplus_query_4031724.sql
@@ -5,8 +5,7 @@
 -- Parameters:
 --  {{start_time}} - the start date timestamp for the accounting period  (inclusively)
 --  {{end_time}} - the end date timestamp for the accounting period (exclusively)
--- {{blockchain}} -- the chain we are interested in
--- {{budget}} -- the amount of COW that needs to be distributed
+-- {{cow_budget}} -- the amount of COW that needs to be distributed
 
 with cow_amm_surplus as (
     select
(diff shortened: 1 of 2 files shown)


cowamm/cow_amm_cow_with_users_surplus_query_4031724.sql
```diff
@@ -19,38 +18,92 @@ with cow_amm_surplus as (
...
     select
         arps.solver_name,
         total_cow_surplus_in_usd,
-        {{budget}} * arps.total_cow_surplus_in_usd / ts.total_surplus_in_usd as total_cow_reward
-    from aggregate_results_per_solver as arps cross join total_surplus as ts
+        {{cow_budget}} * arps.total_cow_surplus_in_usd / ts.total_surplus_in_usd as total_cow_reward
+    from aggregate_results_per_solver_all_chains as arps cross join total_surplus as ts
+),
+
+named_results as (
```
fleupold, line 94: nit

```suggestion
reward_targets as (
```

is clearer imo (generally there are a lot of little meaning table names in this query like temp, final, results etc)

cowamm/cow_amm_cow_with_users_surplus_query_4031724.sql
```diff
@@ -1,12 +1,11 @@
 -- This query computes how much surplus has been provided to CoW AMMs, when trading with other user orders
```
fleupold, line 1: Can we move this file into a bounty subfolder to keep the main cow amm folder lean?

cowamm/cow_amm_cow_with_users_surplus_query_4031724.sql
```diff
@@ -19,38 +18,92 @@ with cow_amm_surplus as (
     where istrade
 ),
 
-cow_surplus_per_batch as (
+cow_surplus_per_batch_ethereum as (
```
fleupold, line 21: Shouldn't we also add arbitrum?

Maybe we should keep the current query that takes blockchain as an arg and gives us `solver_name`, `total_cow_surplus_in_usd` and then have the first query in this file be something like


```
with 
aggregate_results_per_solver_all_chains as (
    select
        solver_name,
        sum(total_cow_surplus_in_usd) as total_cow_surplus_in_usd
    from (
         select * from queryX(blockchain='ethereum')
         union all
         select * from queryX(blockchain='gnosis')
         union all
         select * from queryX(blockchain='arbitrum')
     )
    group by solver_name
),
```

cowamm/cow_amm_cow_with_users_surplus_query_4031724.sql
```diff
@@ -19,38 +18,92 @@ with cow_amm_surplus as (
...
+),
+
+named_results as (
+    select
+        reward_target,
+        substring(solver_name, 6, 100) as solver_name
+    from "query_1541516(end_time='{{end_time}}',vouch_cte_name='named_results')"
+),
+
+final_results_per_solver as (
```
fleupold, line 101: last cte is not needed and in fact the final select seems to be expecting another table (and not use this query at all)

fleupold:
@harisang what is the status here?

harisang:
Closing as it was replaced by PR #60

PR #53, opened 2024-10-03
Add missing eth flow refunds query

fhenneke:
This adds a query for detecting missing eth flow refunds.

It is a for of [this query](https://dune.com/queries/1944598) with renamed parameters and formatted to follow the standards of this repo.

It would be useful to clean the query up a bit before merging.


cowprotocol/eth_flow/.sqlfluff
```diff
@@ -0,0 +1,4 @@
+[sqlfluff:templater:jinja:context]
```
fleupold, line 1: all other folders are without _ so I think this one should be ethflow to follow the standard

fleupold (approved):
Honestly, I think this query is pretty ok as it is (not sure how to clean it up in a meaningful way and it has been working well)

PR #55, opened 2024-10-18
Add solver approvals query

harisang:
This PR migrates the old solver approvals query -> https://dune.com/queries/1973291
that is not functioning anymore into a new query that is also slightly improved -> https://dune.com/queries/4173928

The main improvements are:

1.  The query now also checks for approvals not set by solvers by using a placeholder NON-SOLVER string instead of address; in such a case, the name of the responsible address is set to "NON-SOLVER".
2. It works for Gnosis Chain and Arbitrum as well

An improvement that could be done is to pull out the hardcoded proposer accounts from the query and have those as a separate "whitelisted addresses" query, that we can invoke from within the query of this PR.

### DIFF
--- a/cowprotocol/monitoring/solver_approvals_4173928.sql
+++ b/cowprotocol/monitoring/solver_approvals_4173928.sql
@@ -17,7 +17,10 @@ select --noqa: ST06
     block_time,
     tx_hash,
     responsible_address,
-    coalesce(concat(environment, '-', name), 'NON-SOLVER') as responsible_solver,
+    case
+        when responsible_address = 0x05C5494572E4aB2d48D3AB3aAF6bD4e7b1c98382 then 'PROPOSER-ACCOUNT'
+        else coalesce(concat(environment, '-', name), 'NON-SOLVER')
+    end as responsible_solver,
     spender,
     token,
     value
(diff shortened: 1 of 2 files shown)


cowprotocol/monitoring/solver_approvals_4173928.sql
```diff
@@ -0,0 +1,30 @@
...
+        spender,
+        contract_address as token,
+        a.value,
+        rank() over (partition by spender, contract_address order by evt_block_number desc, evt_index desc) as rk
+    from erc20_{{blockchain}}.evt_Approval as a
+    inner join {{blockchain}}.transactions on evt_tx_hash = hash
+    where owner = 0x9008D19f58AAbD9eD0D60971565AA8510560ab41
+)
+
+select --noqa: ST06
```
MartinquaXD, line 16: what does this comment mean?

MartinquaXD (approved):

cowprotocol/monitoring/solver_approvals_4173928.sql
fhenneke: It disables a formatting check. The check is that in all select statements, one first has simple column names, then computed columns. We disable it often since deviating ordering of columns makes some queries easier to read.

fhenneke (approved):

PR #58, opened 2024-10-25
Update bonding pools

bram-vdberg:
This PR updates the hardcoded bonding pools to include Rizzolver and Furucombo/Portus. The pools can be found on the CMS under solver-bonding pools

PR #59, opened 2024-10-29
Update reduced bonding pools in service fees query

bram-vdberg:
This PR updates the reduced bonding pools in the service fees query after #58

### DIFF
--- a/cowprotocol/accounting/rewards/mainnet/service_fee_query_4017925.sql
+++ b/cowprotocol/accounting/rewards/mainnet/service_fee_query_4017925.sql
@@ -27,43 +27,43 @@ colocated_solvers as (
         'prod-Rizzolver' as solver_name,
         'Reduced-Bonding' as pool_name,
         timestamp '2024-10-10 02:03:00' as joined_on,
-        from_hex('0x0deb0ae9c4399c51289adb1f3ed83557a56df657') as pool_address,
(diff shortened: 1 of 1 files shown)


harisang (changes requested):
Could you follow the format we use for the rest of the reduced bonding pools? There are additional entries such as solver name, joined_on etc and the pool_name has so far been set to 'Reduced-Bonding'. We can of course reevaluate that but it would need to be uniform over all subpools

harisang:
If you look at the previous entries, we have the exact name with the environment included, e.g, prod-Barter, barn-Barter etc. Not sure if it is really needed anywhere, but could you follow that format to avoid having something break?

bram-vdberg:
I updated the pool names to show prod/barn. I also renamed Furucombo to Portus to be in line with their rebranding.

fleupold (approved):

harisang:
The query is still not uniform and new entries don't follow the already used naming conventions.

image

harisang:
@bram-vdberg Took the initiative to push a commit myself as i believe this is currently causing issues with today's payouts (since we switched colocated Rizzolver on yesterday). Could you have a quick look? Thanks!

bram-vdberg (commented):
LGTM, I ran the query on Dune as a test: https://dune.com/queries/4246690. I noticed prod-Portus and prod-Fractal are listed even though they're not live yet, I assume that's not a problem.

harisang (approved):

PR #60, opened 2024-11-05
Fix CoW AMM CoW surplus query

fhenneke:
This PR updates the query for computing CoW surplus of CoW AMMs for the onboing bounty.

The old query was broken due to 80/20-pools since it used a formula for surplus which worked directly with reserves and product invariant ("non-linear surplus"). The resulted in the potential of negative surplus, messing up bounty rewards:

image

The new query uses surplus as defined in the competition ("linear suplus"). This generalizes to all kinds of CoW AMMs. It also reduces the dependency on a large table which indexes reserves of all CoW AMMs. The results with the updated query ([test query](https://dune.com/queries/4243577?end_time_d83555=2024-11-05+00%3A00%3A00&start_time_d83555=2024-10-29+00%3A00%3A00)) change to this:

image

One change in behavior is that we now do not reward solvers in cases where the limit price set by the solver does not lie on the curve. This is, however, in line with how we treat rewards in general and should not have a major impact.

### DIFF
--- a/cowamm/cow_amm_cow_with_users_surplus_query_4031724.sql
+++ b/cowamm/cow_amm_cow_with_users_surplus_query_4031724.sql
@@ -39,10 +39,28 @@ cow_surplus_per_batch_gnosis as (
     where trades.trader in (select address from query_3959044 where blockchain = 'gnosis')
 ),
 
+cow_surplus_per_batch_arbitrum as (
+    select
+        'arbitrum' as blockchain,
+        cow_per_batch.block_time,
+        cow_per_batch.tx_hash,
+        solvers.name as solver_name,
+        naive_cow,  -- fraction of batch volume traded within a CoW
+        trades.surplus_usd as surplus_in_usd,  -- surplus of the executed CoW AMM order, expressed in USD
+        naive_cow * trades.surplus_usd as realized_cow_surplus_in_usd -- surplus of the CoW AMM that is assumed to be generated via a CoW.
+    from "query_4025739(blockchain='arbitrum',start_time='{{start_time}}',end_time='{{end_time}}')" as cow_per_batch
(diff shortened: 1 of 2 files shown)


cowamm/cow_amm_cow_with_users_surplus_query_4031724.sql
```diff
@@ -1,56 +1,78 @@
...
+        trades.surplus_usd as surplus_in_usd,  -- surplus of the executed CoW AMM order, expressed in USD
+        naive_cow * trades.surplus_usd as realized_cow_surplus_in_usd -- surplus of the CoW AMM that is assumed to be generated via a CoW.
+    from "query_4025739(blockchain='gnosis',start_time='{{start_time}}',end_time='{{end_time}}')" as cow_per_batch
+    inner join cow_protocol_gnosis.trades as trades on cow_per_batch.tx_hash = trades.tx_hash
+    inner join cow_protocol_gnosis.batches as batches on cow_per_batch.tx_hash = batches.tx_hash
+    inner join cow_protocol_gnosis.solvers as solvers on batches.solver_address = solvers.address and solvers.active
+    where trades.trader in (select address from query_3959044 where blockchain = 'gnosis')
+),
+
+cow_surplus_per_batch as (
```
fleupold, line 42: Why are we not including arbitrum AMMs here? I think they also qualify for the bounty.

fleupold (approved):

cowamm/cow_amm_cow_with_users_surplus_query_4031724.sql
fhenneke: I added a table for arbitrum. Interestingly, there is no difference in total surplus. I briefly checked the intermediate tables and they seem to work. So I think it is fine to merge this.

PR #61, opened 2024-11-06
Comptetitor AMMs KPIs

PoloX2021:
This PR measures TVL, 24H Volume and 24H APR for 4 existing AMMs (Uniswap, Sushiswap, Pancakeswap and curve)
The 24H APR is defined as `sum(volume*feePct/TVL)*365` over the past 24H.

Uni, Sushi and Pancake are based on the same code. The queries are very similar to one another with only differences in the tables to look into.
For each, there is one first query which looks at the reserves of each pool, to find the largest ones. 
Then the query sync comes from previous code and kpis is a modified version of the 10k comparison to measure only the 24H APR.

For Curve, there are less tables available and information in the events. So, we calculate the reserves based on all transfers of the 2 tokens involved in the pools. 
The trades are dervived from a dedicated dune table.
There might be one issue with ETH transfer which seems to impact only one pool with appears to have a negative TVL.

### DIFF
--- a/cowamm/profitability/competitor_kpis/curve/curve_kpis_4232873.sql
+++ b/cowamm/profitability/competitor_kpis/curve/curve_kpis_4232873.sql
@@ -1,0 +1,25 @@
+-- Computes volume, tvl and APR for Curve pools
+-- APR is measured as the fees earned per $ invested, over the last 24 hours, projected over 1 year
+-- Input: blockchain
+select r.contract_address,
+sum(amount_usd) as volume,
+365*greatest(0,sum(amount_usd*fee/tvl)) as apr,
+avg(fee) as fee,
+avg(tvl) as tvl
+from "query_4232976(blockchain='{{blockchain}}')" r
+left join curve.trades t 
+on r.contract_address = t.project_contract_address
+and r.tx_hash = t.tx_hash
+where t.block_time>=date_add('day', -1, now())
+group by r.contract_address
+
+union
+select 
+    contract_address, 
+    0 as volume, 
+    0 as apr,
+    fee,
+    tvl
+from "query_4232976(blockchain='{{blockchain}}')"
+where time<date_add('day', -1, now())
+and latest = 1
--- a/cowamm/profitability/competitor_kpis/curve/curve_kpis_4232873.sql
+++ b/cowamm/profitability/competitor_kpis/curve/curve_kpis_4232873.sql
@@ -3,7 +3,7 @@
 -- Input: blockchain
 select r.contract_address,
 sum(amount_usd) as volume,
-365*greatest(0,sum(amount_usd*fee/tvl)) as apr,
+365*sum(amount_usd*fee/tvl) as apr,
 avg(fee) as fee,
 avg(tvl) as tvl
 from "query_4232976(blockchain='{{blockchain}}')" r
@@ -11,6 +11,8 @@ left join curve.trades t
 on r.contract_address = t.project_contract_address
 and r.tx_hash = t.tx_hash
 where t.block_time>=date_add('day', -1, now())
+-- This test avoids any possible issue with reconstructing the reserves of the pool
+and tvl >0
 group by r.contract_address
 
 union
--- a/cowamm/profitability/competitor_kpis/curve/curve_largest_2token_pools_4232976.sql
+++ b/cowamm/profitability/competitor_kpis/curve/curve_largest_2token_pools_4232976.sql
@@ -1,0 +1,66 @@
+-- Finds all the curve pools with 2 tokens and their TVLs
+-- Input: blockchain
+
+with 
+-- filters pools with 2 tokens
+pools as(
+select pool_address as contract_address,
+coin0 as token0,
+coin1 as token1,
+mid_fee*power(10,-10) as fee
+from curvefi_{{blockchain}}.view_pools
+where coin2 = 0x0000000000000000000000000000000000000000),
+
+-- finds all transfers in and out of the pools to rebuild the reserves
+transfers as(
+select p.contract_address, token0, token1, -value as transfer0, 0 as transfer1, evt_index, evt_tx_hash as tx_hash, evt_block_time as time, fee
+    from pools p
+    join erc20_{{blockchain}}.evt_transfer t0
+    on p.contract_address = t0."from"
+    and p.token0 = t0.contract_address
+union
+select p.contract_address, token0, token1, value as transfer0, 0 as transfer1, evt_index, evt_tx_hash as tx_hash, evt_block_time as time, fee
+    from pools p
+    join erc20_{{blockchain}}.evt_transfer t0
+    on p.contract_address = t0.to
+    and p.token0 = t0.contract_address
+union
+select p.contract_address, token0, token1, 0 as transfer0, -value as transfer1, evt_index, evt_tx_hash as tx_hash, evt_block_time as time, fee
+    from pools p
+    join erc20_{{blockchain}}.evt_transfer t1
+    on p.contract_address = t1."from"
+    and p.token1 = t1.contract_address
+union
+select p.contract_address, token0, token1, 0 as transfer0, value as transfer1, evt_index, evt_tx_hash as tx_hash, evt_block_time as time, fee
+    from pools p
+    join erc20_{{blockchain}}.evt_transfer t1
+    on p.contract_address = t1.to
+    and p.token1 = t1.contract_address),
+
+-- rebuilds the reserves from the transfers
+-- ETH transfers are not considered  
+reserves as(
+select contract_address, token0, token1, tx_hash, time,
+sum(transfer0) over (partition by contract_address order by time, evt_index) as reserve0,
+sum(transfer1) over (partition by contract_address order by time, evt_index) as reserve1,
+fee,
+ROW_NUMBER() OVER (PARTITION BY tx_hash,contract_address ORDER BY evt_index DESC) AS row_num,
+ROW_NUMBER() OVER (PARTITION BY contract_address ORDER BY time DESC) AS latest
+from transfers)
+
+-- finds the TVL of the pools
+select r.contract_address, token0, token1, time, tx_hash,
+reserve0, reserve1,
+(reserve0 * p0.price / pow(10, p0.decimals)) + (reserve1 * p1.price / pow(10, p1.decimals)) as tvl,
+fee, latest
+from reserves r
+inner join prices.usd as p0
+        on
+            date_trunc('minute', time) = p0.minute
+            and token0 = p0.contract_address
+    inner join prices.usd as p1
+        on
+            date_trunc('minute', time) = p1.minute
+            and token1 = p1.contract_address 
+where row_num = 1
+
--- a/cowamm/profitability/competitor_kpis/pancakeswap/pancakeswap_kpis_4232738.sql
+++ b/cowamm/profitability/competitor_kpis/pancakeswap/pancakeswap_kpis_4232738.sql
@@ -1,0 +1,79 @@
+-- Computes volume, tvl and APR for Pancakeswap pools
+-- APR is measured as the fees earned per $ invested, over the last 24 hours, projected over 1 year
+-- Input: blockchain
+
+with 
+-- select the pool with the largest latest k
+pool as (
+    select
+        contract_address,
+        token0,
+        token1
+    from "query_4232660(blockchain='{{blockchain}}')"
+    where latest = 1
+),
+
+swaps as (
+    select
+        tx_hash as evt_tx_hash,
+        index as evt_index,
+        block_time as evt_block_time,
+        block_number as evt_block_number,
+        contract_address,
+        varbinary_to_uint256(substr(data, 1, 32)) as amount0In,
+        varbinary_to_uint256(substr(data, 33, 32)) as amount1In,
+        varbinary_to_uint256(substr(data, 65, 32)) as amount0Out,
+        varbinary_to_uint256(substr(data, 97, 32)) as amount1Out
+    from {{blockchain}}.logs
+    where
+        topic0 = 0xd78ad95fa46c994b6551d0da85fc275fe613ce37657fb8d5e3d130840159d822 -- Swap
+        and contract_address in (select contract_address from pool)
+),
+
+-- gets the swapped volume and tvl at the time of the swap for each swap
+tvl_volume_per_swap as (
+    select
+        syncs.contract_address as contract_address,
+        syncs.evt_block_time,
+        syncs.evt_tx_hash,
+        (amount0In * p0.price / pow(10, p0.decimals)) + (amount1In * p1.price / pow(10, p1.decimals)) as volume_in,
+        (amount0Out * p0.price / pow(10, p0.decimals)) + (amount1Out * p1.price / pow(10, p1.decimals)) as volume_out,
+        (reserve0 * p0.price / pow(10, p0.decimals)) + (reserve1 * p1.price / pow(10, p1.decimals)) as tvl
+    from "query_4232660(blockchain='{{blockchain}}')" as syncs
+    inner join swaps
+        on
+            syncs.evt_tx_hash = swaps.evt_tx_hash
+            and syncs.contract_address = swaps.contract_address
+            and syncs.evt_index + 1 = swaps.evt_index
+    inner join pool
+        on syncs.contract_address = pool.contract_address
+    inner join prices.usd as p0
+        on
+            date_trunc('minute', syncs.evt_block_time) = p0.minute
+            and syncs.token0 = p0.contract_address
+    inner join prices.usd as p1
+        on
+            date_trunc('minute', syncs.evt_block_time) = p1.minute
+            and syncs.token1 = p1.contract_address
+    where syncs.evt_block_time >= date(date_add('day', -7, now()))
+)
+
+select
+    contract_address,
+    sum((volume_in + volume_out) / 2) as volume,
+    avg(tvl) as tvl,
+    365*sum((volume_in + volume_out) / 2 / tvl) * 0.003 as apr,
+    0.003 as fee
+from tvl_volume_per_swap
+    where evt_block_time >= date_add('day', -1, now())
+group by contract_address
+
+union
+select 
+    pool_address as contract_address, 
+    0 as volume, 
+    tvl,
+    0 as apr,
+    0.003 as fee
+from "query_4232597(blockchain='{{blockchain}}')"
+where pool_address not in (select contract_address from tvl_volume_per_swap)
--- a/cowamm/profitability/competitor_kpis/pancakeswap/pancakeswap_largest_pools_4232597.sql
+++ b/cowamm/profitability/competitor_kpis/pancakeswap/pancakeswap_largest_pools_4232597.sql
@@ -1,0 +1,49 @@
+-- Computes the TVL for every Pancakeswap pool
+-- Then returns the top {{number_of_pools}} pools by TVL
+-- Input: blockchain, number_of_pools to return
+
+with 
+-- finds the pools which have been active since 2024-10-01
+data as (
+select 
+    p.contract_address as pool_address,
+    call_block_time as time,
+    output__reserve0 as balance0,
+    output__reserve1 as balance1,
+    rank() over (partition by p.contract_address order by p.call_block_time desc) latest
+from pancakeswap_v2_{{blockchain}}.PancakePair_call_getReserves p 
+where date_trunc('day',call_block_time) >= date_trunc('day', cast('2024-10-01' as date))),
+
+--Gets the token0 and token1 addresses for each pool
+t0 as (select contract_address, max(output_0) as token0 
+    from pancakeswap_v2_{{blockchain}}.PancakePair_call_token0
+    group by contract_address),
+    
+t1 as (select contract_address, max(output_0) as token1
+    from pancakeswap_v2_{{blockchain}}.PancakePair_call_token1
+    group by contract_address),
+
+--computes the tvl for each pool
+-- for each pool we could get multiple balance values if the function was called multiple times in a same block
+-- we arbitrarily choose the maximum value for each pool
+recent_tvl as(
+select pool_address,
+token0, max(balance0) as balance0,
+token1, max(balance1) as balance1,
+max(least(balance0,balance1)* greatest(p0.price/pow(10, p0.decimals),p1.price/pow(10, p1.decimals)) +
+ greatest(balance0,balance1)* least(p0.price/pow(10, p0.decimals),p1.price/pow(10, p1.decimals))) as tvl
+from data
+join t0
+    on pool_address = t0.contract_address
+join t1
+    on pool_address = t1.contract_address
+join prices.usd_latest as p0
+    on token0 = p0.contract_address
+join prices.usd_latest as p1
+    on token1 = p1.contract_address
+where latest = 1
+group by 1,2,4)
+
+select * from recent_tvl
+order by tvl desc
+limit {{number_of_pools}}
--- a/cowamm/profitability/competitor_kpis/pancakeswap/pancakeswap_syncs_4232660.sql
+++ b/cowamm/profitability/competitor_kpis/pancakeswap/pancakeswap_syncs_4232660.sql
@@ -1,0 +1,27 @@
+-- Finds the pancakeswap v2 pool address given tokens specified in query parameters (regardless of order)
+with pools as (
+    select
+        substr(data, 13, 20) as contract_address,
+        substr(topic1, 13, 20) as token0,
+        substr(topic2, 13, 20) as token1
+    from {{blockchain}}.logs
+    where
+        topic0 = 0x0d3648bd0f6ba80134a33ba9275ac585d9d315f0ad8355cddefde31afa28d0e9 -- PairCreated
+        -- topic1: 0x0...0<token0>, topic2: 0x0...0<token1>
+        and substr(data, 13, 20) in (select pool_address from "query_4232597(blockchain='{{blockchain}}')")
+)
+
+select
+    pools.*,
+    tx_hash as evt_tx_hash,
+    index as evt_index,
+    block_time as evt_block_time,
+    block_number as evt_block_number,
+    varbinary_to_uint256(substr(data, 1, 32)) as reserve0,
+    varbinary_to_uint256(substr(data, 33, 32)) as reserve1,
+    rank() over (partition by (logs.contract_address) order by block_time desc) as latest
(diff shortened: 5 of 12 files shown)


cowamm/profitability/competitor_kpis/curve/curve_kpis_4232873.sql
```diff
@@ -0,0 +1,31 @@
+-- Computes volume, tvl and APR for Curve pools
+-- APR is measured as the fees earned per $ invested, over the last 24 hours, projected over 1 year
+-- Input: blockchain
+select
+    r.contract_address,
+    sum(amount_usd) as volume,
+    365 * sum(amount_usd * fee / tvl) as apr,
+    avg(fee) as fee,
+    avg(tvl) as tvl
```
fleupold, line 9: Are we do we want the avg fee and tvl here (this is a number that may have never existed as a true TVL)? I personally think the latest values for fee and tvl would make more sense.

cowamm/profitability/competitor_kpis/curve/curve_kpis_4232873.sql
```diff
@@ -0,0 +1,31 @@
...
+select
+    contract_address,
+    0 as volume,
+    0 as apr,
+    fee,
+    tvl
+from "query_4232976(blockchain='{{blockchain}}')"
+where
+    time < date_add('day', -1, now())
+    and latest = 1
```
fleupold, line 31: What does this do? Is this for the case that tvl is 0? I think it would be easier to use an if statement in the apr computation line (or omit these pools altogether as they seem irrelevant).

cowamm/profitability/competitor_kpis/curve/curve_largest_2token_pools_4232976.sql
```diff
@@ -0,0 +1,65 @@
...
+pools as(
+select pool_address as contract_address,
+coin0 as token0,
+coin1 as token1,
+mid_fee*power(10,-10) as fee
+from curvefi_{{blockchain}}.view_pools
+where coin2 = 0x0000000000000000000000000000000000000000),
+
+-- finds all transfers in and out of the pools to rebuild the reserves
+transfers as(
```
fleupold, line 15: Can't re replace the four unions by something like https://github.com/cowprotocol/dune-queries/blob/b26d06e362b2c4a6c17740d42181d88cbd872670/cowamm/kpi/tvl_by_chain_4096107.sql#L25-L40 which I believe accomplishes the same thing in half the lines

cowamm/profitability/competitor_kpis/curve/curve_largest_2token_pools_4232976.sql
```diff
@@ -0,0 +1,65 @@
...
+
+with 
+-- filters pools with 2 tokens
+pools as(
+select pool_address as contract_address,
+coin0 as token0,
+coin1 as token1,
+mid_fee*power(10,-10) as fee
+from curvefi_{{blockchain}}.view_pools
+where coin2 = 0x0000000000000000000000000000000000000000),
```
fleupold, line 12: Why is this needed? Intuitively this looks to me as if this is an invalid pool (with coin2 being the 0 address). Could you add a comment in code please?

cowamm/profitability/competitor_kpis/pancakeswap/pancakeswap_largest_pools_4232597.sql
```diff
@@ -0,0 +1,64 @@
+-- Computes the TVL for every Pancakeswap pool
+-- Then returns the top {{number_of_pools}} pools by TVL
```
fleupold, line 2: Why does curve not have `number_of_pools` (can we make it consistent) and also it seems this value is never set.

cowamm/profitability/competitor_kpis/pancakeswap/pancakeswap_largest_pools_4232597.sql
```diff
@@ -0,0 +1,64 @@
...
+with
+-- finds the pools which have been active since 2024-10-01
+data as (
+    select
+        p.contract_address as pool_address,
+        p.call_block_time as time,
+        p.output__reserve0 as balance0,
+        p.output__reserve1 as balance1,
+        rank() over (partition by p.contract_address order by p.call_block_time desc) as latest
+    from pancakeswap_v2_{{blockchain}}.PancakePair_call_getReserves as p
```
fleupold, line 14: Is there not a better way to get this data. `_call_` tables are populated if a view function is invoked as part of a transaction, but this doesn't always fave to be the case. This was a source of inaccuracy for the initial CoW AMM performance queries. In order to get reserves we should also build a complete transfer ledger as we do for CoW AMM (unless there is already a persisted Dune view that accomplishes this)

cowamm/profitability/competitor_kpis/pancakeswap/pancakeswap_largest_pools_4232597.sql
```diff
@@ -0,0 +1,64 @@
...
+    select
+        contract_address,
+        max(output_0) as token1
+    from pancakeswap_v2_{{blockchain}}.PancakePair_call_token1
+    group by contract_address
+),
+
+--computes the tvl for each pool
+-- for each pool we could get multiple balance values if the function was called multiple times in a same block
+-- we arbitrarily choose the maximum value for each pool
```
fleupold, line 37: This seems indeed quite arbitrary. Shouldn't we build a ledger with tvl per transaction and then simply pick the most recent one? I would like for our logic to be consistent across pools as this will require a future maintainer to understand less subtleties.

cowamm/profitability/competitor_kpis/pancakeswap/pancakeswap_kpis_4232738.sql
```diff
@@ -0,0 +1,79 @@
...
+    contract_address,
+    sum((volume_in + volume_out) / 2) as volume,
+    avg(tvl) as tvl,
+    365*sum((volume_in + volume_out) / 2 / tvl) * 0.003 as apr,
+    0.003 as fee
+from tvl_volume_per_swap
+    where evt_block_time >= date_add('day', -1, now())
+group by contract_address
+
+union
```
fleupold, line 71: again, I don't understand the point of this union. Would appreciate if it can be simplified away or at least commented in code for future maintainers.

cowamm/profitability/competitor_kpis/pancakeswap/pancakeswap_syncs_4232660.sql
```diff
@@ -0,0 +1,27 @@
+-- Finds the pancakeswap v2 pool address given tokens specified in query parameters (regardless of order)
```
fleupold, line 1: Can we please use consistent structure for all queries (ie the same way of specifying what parameters are needed and why as in the other queries in the repo)?

cowamm/profitability/competitor_kpis/sushiswap/sushiswap_kpis_4232588.sql
```diff
@@ -0,0 +1,79 @@
+-- Computes volume, tvl and APR for Sushiswap pools
+-- APR is measured as the fees earned per $ invested, over the last 24 hours, projected over 1 year
+-- Input: blockchain
+
+with
```
fleupold, line 5: If this queries are 99% the same (maybe modulo some event topic) we should not duplicate all this code, but instead have the thing that changes a parameter (this way the actual pancake, uni, sushi query become just a `select * from query...` and the whole thing will be much easier to maintain.

fleupold (commented):
Can you please run `sqlfluff fix` to make sure the sql is formatted nicely (this makes it easier for me to review). Also, I'd personally prefer smaller PRs (e.g. one per project) as then I can review them 1 by 1 in 20 minutes each (rather than having to block 1h for the whole PR, which at least for me happens less frequently).

As for the structure, the /cowamm currently contains two folders (kip and profitability). Inside profitibality, we now add competitor_kpis. I find this confusing.

Can we either create /cowamm/competitors/kpi or /comamm/kpi/competitors?

More comments (mainly about semantic inconsistencies inline).

cowamm/profitability/competitor_kpis/curve/curve_largest_2token_pools_4232976.sql
PoloX2021: Curve can have multi-tokens pools, which would not make sense to compare with CoW AMM for now

cowamm/profitability/competitor_kpis/curve/curve_kpis_4232873.sql
PoloX2021: This for the case when the pool has not traed in the period specified by the query. It could be interesting to show that their APR is 0% to promote our pools instead.
I thought it was better to implement it in the query itself, than handling this edge case elsewhere.

cowamm/profitability/competitor_kpis/curve/curve_kpis_4232873.sql
PoloX2021: The avg(tvl) was to remain consistent with the other pools where the displayed tvl is averaged over the past period. 
This comes form query_4059213. 
In my opinion, it does make sense, as it removes the noise in the tvl generated by small variations.

cowamm/profitability/competitor_kpis/pancakeswap/pancakeswap_largest_pools_4232597.sql
PoloX2021: Same explanation as previous comment. 
We could worry about recreating an exact value for each AMM but the code and the execution would be much longer.

cowamm/profitability/competitor_kpis/pancakeswap/pancakeswap_kpis_4232738.sql
PoloX2021: Same explanation as first remark. I commented in the code, but can remove it altogether, if it is not the best way to proceed.

cowamm/profitability/competitor_kpis/curve/curve_kpis_4232873.sql
fleupold: But don't you get those with the upper query as well (since it's a left join)? The only issue I see is that you may have a "divide by 0" issue in `sum(amount_usd * fee / tvl)` but there you could use something like [`try`](https://prestodb.io/docs/current/functions/conditional.html#id4) of if statements instead of needed an entire new union query.

cowamm/profitability/competitor_kpis/pancakeswap/pancakeswap_largest_pools_4232597.sql
fleupold: How big is the performance penalty compared to building a proper tvl ledger? I worry that e.g. for pools with on large LP when they remove all of their liquidity the pool stays indexed with their old large tvl value.

If we use a method that we know is inaccurate. let's very clearly document this in the code. Otherwise it just looks as if we think relying on these `_call_` tables is ok (which it generally isn't)

cowamm/profitability/competitor_kpis/curve/curve_kpis_4232873.sql
PoloX2021: Yes, indeed the left join does the work. I removed this part.

cowamm/profitability/competitor_kpis/pancakeswap/pancakeswap_kpis_4232738.sql
PoloX2021: This union seems more useful than for curve pools.
To remove it, it would require to restructure the query. Not sure if it would become clearer.

cowamm/kpi/competitors/curve/curve_largest_2token_pools_4232976.sql
```diff
@@ -0,0 +1,97 @@
...
+        contract_address,
+        token0,
+        token1,
+        tx_hash,
+        block_time,
+        fee,
+        sum(transfer0) over (partition by contract_address order by block_time, evt_index) as reserve0,
+        sum(transfer1) over (partition by contract_address order by block_time, evt_index) as reserve1,
+        row_number() over (partition by tx_hash, contract_address order by evt_index desc) as row_num,
+        row_number() over (partition by contract_address order by block_time desc) as latest
```
fleupold, line 60: I don't understand the difference between the two. row_num is partitioned by tx_hash so it doesn't look at all at the time ordering of the transactions, does it (it just gives the latest reserve per tx)?

Isn't all we need `row_number() over (partition by contract_address order by block_time, evt_index desc)` 
 so that we can get the most up to date reserves?

cowamm/kpi/competitors/uni_swap_style/largest_uni_style_kpis_4304295.sql
```diff
@@ -0,0 +1,75 @@
+-- Computes volume, tvl and APR for Uni Swap style pools (Uni, Pancake, Sushi)
+-- APR is measured as the fees earned per $ invested, over the last 24 hours, projected over 1 year
+-- The type of pool (Uni/Pancake/Sushi) is not specified
+-- Parameters:
+-- {{blockchain}}: The blockchain to query
```
fleupold, line 5: missing `number_of_pools`

cowamm/kpi/competitors/uni_swap_style/largest_uni_style_syncs_4304212.sql
```diff
@@ -0,0 +1,28 @@
+-- This query finds all the sync event which happens after a swap and reference the new reserves of the pool
+-- Parameters:
+-- {{blockchain}}: The blockchain to query
```
fleupold, line 3: number of pools missing

cowamm/kpi/competitors/uni_swap_style/largest_uni_style_kpis_4304295.sql
```diff
@@ -0,0 +1,75 @@
+-- Computes volume, tvl and APR for Uni Swap style pools (Uni, Pancake, Sushi)
+-- APR is measured as the fees earned per $ invested, over the last 24 hours, projected over 1 year
+-- The type of pool (Uni/Pancake/Sushi) is not specified
```
fleupold, line 3: shouldn't we allow for this though (we might now based on the factory contract that emitted the event). I'm pretty sure we need to know which type of pool we are talking about in order to properly render competitor information in the UI.

Ideally the result would contain the amm type or allow filtering for a specific amm type (ie. by passing in the correct factory contract address)

cowamm/kpi/competitors/uni_swap_style/largest_uni_style_kpis_4304295.sql
```diff
@@ -0,0 +1,75 @@
...
+        on
+            date_trunc('minute', syncs.evt_block_time) = p0.minute
+            and syncs.token0 = p0.contract_address
+    inner join prices.usd as p1
+        on
+            date_trunc('minute', syncs.evt_block_time) = p1.minute
+            and syncs.token1 = p1.contract_address
+    where
+        syncs.evt_block_time >= date(date_add('day', -1, now()))
+        and swaps.evt_block_time >= date(date_add('day', -1, now()))
```
fleupold, line 61: It's probably better to pass in the day for which we want to run this query as an argument. This will make it later easier to backfill historic data in case the pipeline fails for a given day

cowamm/kpi/competitors/uni_swap_style/largest_uni_style_syncs_4304212.sql
```diff
@@ -0,0 +1,28 @@
...
+-- Finds the pools with the largest tvl
+with pools as (
+    select
+        pool_address,
+        token0,
+        token1
+    from "query_4303563(blockchain='{{blockchain}}', number_of_pools = '{{number_of_pools}}')"
+)
+
+select
```
fleupold, line 14: Is this query useful as its own standalone query? It looks like it's only used in one place and there we do a bunch of the joins we also do here, so it seems like it might be easier to inline it there?

fleupold (commented):
Looking very good. Would like to see a way to distinguish between different uni v2 like protocols though as this is very likely something the UI wants to display.

cowamm/kpi/competitors/curve/curve_largest_2token_pools_4232976.sql
PoloX2021: The query not only finds the largest pools, but also returns the tvl and the transfers for each tx. This avoids resimulating the reserves multiple times. I'll update the description of the query with this in mind.

latest indicates if this is the most recent reserves to get the current largest pools.
row_num gets the reserves of the pool after each block to compute the tvl at every moment in time

cowamm/kpi/competitors/uni_swap_style/largest_uni_style_kpis_4304295.sql
```diff
@@ -0,0 +1,75 @@
...
+)
+
+select
+    contract_address,
+    0.003 as fee,
```
shoom3301, line 66: Fee is hardcoded to 0.003 here, but some pools might have different fee values, don't they?

PoloX2021: Thanks for the comment! Pancake Swap is only .25%
For Sushi and Uni swap it is harcoded to .3%.

cowamm/kpi/competitors/curve/curve_kpis_4232873.sql
```diff
@@ -0,0 +1,25 @@
...
+-- {{blockchain}}: The blockchain to query
+-- {{end_time}}: The end time of the time window (end_time - 1 day; end_time), defaults to now()
+select
+    r1.contract_address,
+    r1.fee,
+    r1.tvl,
+    sum(amount_usd) as volume,
+    365 * sum(amount_usd * r.fee / r.tvl) as apr
+from "query_4232976(blockchain='{{blockchain}}')" as r
+inner join "query_4232976(blockchain='{{blockchain}}')" as r1
```
fleupold, line 13: Why is this self join needed? Can you please document it in the code.

cowamm/kpi/competitors/curve/curve_kpis_4232873.sql
```diff
@@ -0,0 +1,25 @@
...
+    on
+        r.contract_address = r1.contract_address
+left join curve.trades as t
+    on
+        r.contract_address = t.project_contract_address
+        and r.tx_hash = t.tx_hash
+where
+    t.block_time >= date_add('day', -1, (case when '{{end_time}}' = '2024-01-01' then now() else timestamp '{{end_time}}' end))
+    and r1.latest = 1
```
fleupold, line 22: 4232976 already filters for `latest = 1` doesn't it?

cowamm/kpi/competitors/curve/curve_largest_2token_pools_4232976.sql
fleupold: I see, that's smart! Just a nit but would you mind naming this column a bit more descriptive (ie. `latest_per_tx` & `latest_per_pool`)

cowamm/kpi/competitors/uni_swap_style/largest_uni_style_pools_4303563.sql
```diff
@@ -0,0 +1,64 @@
...
+inner join prices.usd as p0
+    on
+        token0 = p0.contract_address
+        and p0.minute = evt_block_time
+inner join prices.usd as p1
+    on
+        token1 = p1.contract_address
+        and p1.minute = evt_block_time
+where latest = 1
+order by 7 desc
```
fleupold, line 63: If I'm not miscounting this is sorting by evt_block_time not tvl (maybe best to name the column explicitly in case we add more columns)

cowamm/kpi/competitors/uni_swap_style/largest_uni_style_kpis_4304295.sql
```diff
@@ -0,0 +1,101 @@
...
+        block_time as evt_block_time,
+        block_number as evt_block_number,
+        varbinary_to_uint256(substr(data, 1, 32)) as reserve0,
+        varbinary_to_uint256(substr(data, 33, 32)) as reserve1,
+        rank() over (partition by (logs.contract_address) order by block_time desc) as latest
+    from {{blockchain}}.logs
+    inner join pool
+        on logs.contract_address = pool.contract_address
+    where
+        block_time >= date_add('day', -1, (case when '{{end_time}}' = '2024-01-01' then now() else timestamp '{{end_time}}' end))
```
fleupold, line 34: Here we should just use a default value that is very far in the future and not a date that we may actually want to query

cowamm/kpi/competitors/uni_swap_style/largest_uni_style_kpis_4304295.sql
```diff
@@ -0,0 +1,101 @@
+-- Computes volume, tvl and APR for Uni Swap style pools (Uni, Pancake, Sushi)
+-- APR is measured as the fees earned per $ invested, over the last 24 hours, projected over 1 year
+-- The type of pool (Uni/Pancake/Sushi) is not specified
```
fleupold, line 3: Let's remove this comment now (I think the pool is actually specified now)

fleupold (commented):
Very close. I just don't understand the self join in the curve query. The rest are nits.

cowamm/kpi/competitors/curve/curve_kpis_4232873.sql
PoloX2021: No, it might feel confusing in the last part of the code: it filters by latest_per_pool only to find the largest pools, then it includes all of the tvls after every event which happened

`select * from recent_tvl
where contract_address in (
    select contract_address
    from recent_tvl
    where latest_per_pool = 1
    order by tvl desc
    limit {{number_of_pools}}
)
`

cowamm/kpi/competitors/all_competitors_4335231.sql
```diff
@@ -0,0 +1,33 @@
+--Groups all competitors in one query
```
fleupold, line 1: nit: document `{{blockchain}}` parameter

cowamm/kpi/competitors/all_competitors_4335231.sql
```diff
@@ -0,0 +1,33 @@
...
+select
+    contract_address,
+    tvl,
+    fee,
+    volume,
+    apr,
+    project,
+    case
+        when '{{blockchain}}' = 'ethereum' then 1
+        when '{{blockchain}}' = 'ethereum' then 100
```
fleupold, line 13:
```suggestion
        when '{{blockchain}}' = 'gnosis' then 100
```

cowamm/kpi/competitors/all_competitors_4335231.sql
```diff
@@ -0,0 +1,33 @@
...
+select
+    contract_address,
+    tvl,
+    fee,
+    volume,
+    apr,
+    'curve' as project,
+    case
+        when '{{blockchain}}' = 'ethereum' then 1
+        when '{{blockchain}}' = 'ethereum' then 100
```
fleupold, line 30:
```suggestion
        when '{{blockchain}}' = 'gnosis' then 100
```

cowamm/kpi/competitors/all_competitors_4335231.sql
fleupold: Maybe add support for base while we are at it?

cowamm/kpi/competitors/curve/curve_kpis_4232873.sql
```diff
@@ -0,0 +1,27 @@
...
+-- APR is measured as the fees earned per $ invested, over the last 24 hours, projected over 1 year
+-- Parameters:
+-- {{blockchain}}: The blockchain to query
+-- {{competitor_end_time}}: The end time of the time window (end_time - 1 day; end_time), defaults to now()
+select
+    r1.contract_address,
+    r1.fee,
+    r1.tvl,
+    sum(amount_usd) as volume,
+    365 * sum(amount_usd * r.fee / r.tvl) as apr
```
fleupold, line 11: Thanks for the explanation, I get it now. Just as a comment (not actionable assuming performance of the join isn't an issue), but this _could_ also be computed using a "rolling sum" which would avoid the self join.

Something like

```
select
    contract_address,
    fee,
    tvl,
    sum(amount_usd) over (partition by contract_address order by latest_per_pool) as volume,
    365 * sum(amount_usd * fee / tvl) over (partition by contract_address order by latest_per_pool) as apr
```

cowamm/kpi/competitors/curve/curve_largest_2token_pools_4232976.sql
```diff
@@ -0,0 +1,99 @@
...
+reserves as (
+    select
+        contract_address,
+        token0,
+        token1,
+        tx_hash,
+        block_time,
+        fee,
+        sum(transfer0) over (partition by contract_address order by block_time, evt_index) as reserve0,
+        sum(transfer1) over (partition by contract_address order by block_time, evt_index) as reserve1,
```
fleupold, line 60: Looking at the performance of the query and the fact that we are already joining with prices for sell and buy token here, I wonder if we wouldn't dramatically speed up this query by also computing the `amount_usd` per tx_hash here instead of computing it in the other query.

cowamm/kpi/competitors/all_competitors_4335231.sql
```diff
@@ -0,0 +1,33 @@
...
+        when '{{blockchain}}' = 'arbitrum' then 42161
+    end as chain_id
+from "query_4304295(blockchain='{{blockchain}}')"
+
+union distinct
+
+--Curve
+select
+    contract_address,
```
fleupold, line 22: surprised you don't need the token addresses as well here (would make sense imo)

fleupold (approved):
👏  Looks great, only nits remaining

cowamm/kpi/competitors/all_competitors_4335231.sql
PoloX2021: Sasha was happy with all the columns in the table...

cowamm/kpi/competitors/curve/curve_largest_2token_pools_4232976.sql
PoloX2021: Not sure if it would speed up or not, because we would have to look for the transaction events and not simply the transfers

cowamm/kpi/competitors/curve/curve_kpis_4232873.sql
PoloX2021: There is indeed a performance issue overall for all queries. So I implemented it. Result: it divides by 2 the execution time

PR #63, opened 2024-11-07
Add raw price feed query

harisang:
This PR adds a view of the mainnet price feed that we currently upload, that is set up in such a way so that it can be used right away with the existing slippage calculations. This means that for every price feed, and each token, we compute an average price for each hour (as this is what we have also been doing with the Dune price feed).

### DIFF
--- a/cowprotocol/accounting/price_feed/raw_price_feed_query_4252674.sql
+++ b/cowprotocol/accounting/price_feed/raw_price_feed_query_4252674.sql
@@ -1,0 +1,41 @@
+with imported_prices as (
+    select
+        token_address,
+        cast(replace(time, 'T', ' ') as timestamp) as time, --noqa: RF04
+        cast(price as double) as price_unit_eth,
+        decimals,
+        source
+    from dune.cowprotocol.dataset_price_feed_ethereum
+),
+
+processed_data as (
+    select -- noqa: ST06
+        token_address,
+        date_trunc('minute', time) as minute, --noqa: RF04
+        decimals,
+        source,
+        avg(price_unit_eth) as price_unit_eth
+    from imported_prices group by 1, 2, 3, 4
+),
+
+processed_data_with_usd_prices as (
+    select -- noqa: ST06
+        pd.token_address,
+        pd.minute,
+        pd.decimals,
+        pd.price_unit_eth,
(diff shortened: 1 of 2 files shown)


cowprotocol/accounting/price_feed/raw_price_feed_query_4252674.sql
```diff
@@ -0,0 +1,41 @@
...
+        pd.decimals,
+        pd.price_unit_eth,
+        pd.price_unit_eth * p.price as price_unit,
+        pd.price_unit_eth * p.price / pow(10, pd.decimals) as price_atom,
+        source
+    from processed_data as pd inner join prices.usd as p on pd.minute = p.minute
+    where p.contract_address = 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2 and blockchain = 'ethereum'
+)
+
+select -- noqa: ST06
```
fhenneke, line 34: Is this table meant as an addition to the two tables `precise_prices` and `intrinsic_prices` in https://github.com/cowprotocol/dune-queries/blob/main/cowprotocol/accounting/slippage/slippage_prices_4064601.sql? Or can it also be used as price source for computing `intrinsic_prices`?

The former would be fine, I guess, for experiments. But it would make it difficult to only rely on this feed in the future since the exchange rate logic of `intrinsic_prices` might not be easy to combine with the feed.

harisang: As a first attempt, i was thinking to basically rename the prices table from here: https://github.com/cowprotocol/dune-queries/blob/5e3d398cd2f614d10da8db1e80cfc09f4695d9d4/cowprotocol/accounting/slippage/slippage_prices_4064601.sql#L85

as dune_prices, and then have the table that the query in this PR returns split into N source_prices table, where N is the number of price feeds we push to Dune.

cowprotocol/accounting/price_feed/raw_price_feed_query_4252674.sql
```diff
@@ -0,0 +1,41 @@
+with imported_prices as (
```
fhenneke, line 1: This does not have any documentation yet

harisang:
> Looks good except for documentation.

Just added

PR #65, opened 2024-11-13
Add multiple price feeds option

harisang:
This PR adds the option to use multiple price feeds as follows:
- at most 4 price feeds are considered, and the` approx_percentile(0.5)` is used to compute a price based on these price feeds.  This replaces what has been now only the dune price.

In case all price feeds return null, we continue to use the fallbacks that have been used up till now

### DIFF
--- a/cowprotocol/accounting/slippage/slippage_prices_4064601.sql
+++ b/cowprotocol/accounting/slippage/slippage_prices_4064601.sql
@@ -25,8 +25,47 @@ with token_times as (
     group by 1, 2
 ),
 
+imported_price_feed as (
+    select *
+    from "query_4252674"
+    where price_unit < 10000000 -- here we filter all tokens with price more than $1M, as these are probably bogus prices
+),
+
+coingecko_price_feed as (
+    select
+        ipf.hour,
+        ipf.token_address,
+        ipf.decimals,
+        ipf.price_unit,
+        ipf.price_atom
+    from imported_price_feed as ipf inner join token_times as tt on ipf.hour = tt.hour and ipf.token_address = tt.token_address
+    where source = 'coingecko'
+),
+
+moralis_price_feed as (
+    select
+        ipf.hour,
+        ipf.token_address,
+        ipf.decimals,
+        ipf.price_unit,
+        ipf.price_atom
+    from imported_price_feed as ipf inner join token_times as tt on ipf.hour = tt.hour and ipf.token_address = tt.token_address
+    where source = 'moralis'
+),
+
+auction_price_feed as (
+    select
+        ipf.hour,
+        ipf.token_address,
+        ipf.decimals,
+        ipf.price_unit,
+        ipf.price_atom
+    from imported_price_feed as ipf inner join token_times as tt on ipf.hour = tt.hour and ipf.token_address = tt.token_address
+    where source = 'native'
+),
+
 -- Precise prices are prices from the Dune price feed.
-precise_prices as (
+dune_price_feed as (
     select -- noqa: ST06
         date_trunc('hour', minute) as hour, --noqa: RF04
         token_address,
@@ -41,6 +80,35 @@ precise_prices as (
             and contract_address = token_address
             and blockchain = '{{blockchain}}'
     group by 1, 2, 3
+),
+
+multiple_price_feeds_pre as (
+    select *
+    from dune_price_feed
+    union all
+    select *
+    from coingecko_price_feed
+    union all
+    select *
+    from moralis_price_feed
+    union all
+    select *
+    from auction_price_feed
+),
+
+multiple_price_feeds as (
+    select
+        hour,
+        token_address,
+        decimals,
+        approx_percentile(price_unit, 0.5) as price_unit,
+        approx_percentile(price_atom, 0.5) as price_atom
+    from multiple_price_feeds_pre group by 1, 2, 3
+),
+
+precise_prices as (
+    select *
+    from {{price_feed}}
 ),
 
 -- Intrinsic prices are prices reconstructed from exchange rates from within the auction
--- a/cowprotocol/accounting/rewards/mainnet/unusual_slippage_query_2332678.sql
+++ b/cowprotocol/accounting/rewards/mainnet/unusual_slippage_query_2332678.sql
@@ -16,6 +16,7 @@ select  --noqa: ST06
     concat(
         '<a href="https://dune.com/queries/4059683',
         '?blockchain=ethereum',
+        '&price_feed=dune_price_feed'
         '&start_time={{start_time}}',
         '&end_time={{end_time}}',
         '&slippage_table_name=raw_slippage_breakdown',
--- a/cowprotocol/accounting/slippage/.sqlfluff
+++ b/cowprotocol/accounting/slippage/.sqlfluff
@@ -4,3 +4,4 @@ end_time='2024-08-02 12:00'
 blockchain='ethereum'
 slippage_table_name=slippage_per_solver,slippage_per_transaction
 raw_slippage_table_name=raw_slippage_breakdown,raw_slippage_per_transaction
+price_feed=dune_price_feed,multiple_price_feeds
--- a/cowprotocol/accounting/slippage/raw_slippage_4059683.sql
+++ b/cowprotocol/accounting/slippage/raw_slippage_4059683.sql
@@ -46,7 +46,7 @@ fees as (
         -amount as amount,
         fee_type as slippage_type,
         date_trunc('hour', block_time) as hour --noqa: RF04
-    from "query_4058574(blockchain='{{blockchain}}',start_time='{{start_time}}',end_time='{{end_time}}')"
+    from "query_4058574(blockchain='{{blockchain}}',price_feed='{{price_feed}}',start_time='{{start_time}}',end_time='{{end_time}}')"
 ),
 
 imbalances as (
--- a/cowprotocol/accounting/slippage/slippage_4070065.sql
+++ b/cowprotocol/accounting/slippage/slippage_4070065.sql
@@ -34,7 +34,7 @@ slippage_per_transaction as (
         solver_address,
         sum(slippage_usd) as slippage_usd,
         sum(slippage_wei) as slippage_wei
-    from "query_4059683(blockchain='{{blockchain}}',start_time='{{start_time}}',end_time='{{end_time}}',raw_slippage_table_name='raw_slippage_breakdown')" as rs
+    from "query_4059683(blockchain='{{blockchain}}',price_feed='{{price_feed}}',start_time='{{start_time}}',end_time='{{end_time}}',raw_slippage_table_name='raw_slippage_breakdown')" as rs
     inner join cow_protocol_{{blockchain}}.batches as b
         on rs.tx_hash = b.tx_hash
     where rs.tx_hash not in (select tx_hash from excluded_batches)


cowprotocol/accounting/slippage/slippage_prices_4064601.sql
```diff
@@ -43,6 +82,35 @@ precise_prices as (
...
+    from coingecko_price_feed
+    union all
+    select *
+    from moralis_price_feed
+    union all
+    select *
+    from auction_price_feed
+),
+
+multiple_price_feeds as (
```
fhenneke, line 99: I would probably rename to `median_price_feed`.

cowprotocol/accounting/slippage/slippage_prices_4064601.sql
```diff
@@ -25,8 +25,47 @@ with token_times as (
     group by 1, 2
 ),
 
+imported_price_feed as (
```
fhenneke, line 28: Please add some comments on what those tables mean, or on the meaning of some tables in the context of the full query.

fhenneke (dismissed):
The changes generally make sense for testing other price feeds.

One issue is that the query is a bit too complicated for me now. (Maybe already was too complicated before.) Since it is essential to the accounting, it has to easy to understand what the query does.

For example, the handling of missing values seems a bit indirect: first, some price feeds are combined and an approximate median is computed. If that is missing, yet another price feed is used. So there seem to be different tiers of price feeds as well. But that is not made explicit anywhere. And only the Dune price feed is used for some things (intrinsic prices, native token prices) that other price feeds could theoretically also be used for.

cowprotocol/accounting/slippage/slippage_prices_4064601.sql
```diff
@@ -25,8 +25,47 @@ with token_times as (
...
+    select *
+    from "query_4252674"
+    where price_unit < 10000000 -- here we filter all tokens with price more than $1M, as these are probably bogus prices
+),
+
+coingecko_price_feed as (
```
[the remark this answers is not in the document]
harisang: It was basically to make it explicit what price feeds we are using for slippage accounting. There is a raw price feed query but this just returns everything in one table. This breakdown also makes it easier to remove a price feed if we want to.

but i agree that in its current form, it doesn't add anything to the query so i can remove it if you think it makes it harder to read

cowprotocol/accounting/slippage/slippage_prices_4064601.sql
```diff
@@ -43,6 +82,35 @@ precise_prices as (
...
+    select *
+    from auction_price_feed
+),
+
+multiple_price_feeds as (
+    select
+        hour,
+        token_address,
+        decimals,
+        approx_percentile(price_unit, 0.5) as price_unit,
```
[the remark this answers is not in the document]
harisang: Switched the code to compute the correct median, based on the No2 solution proposed here
https://medium.com/learning-sql/how-to-calculate-median-the-right-way-in-postgresql-f7b84e9e2df7

cowprotocol/accounting/slippage/slippage_prices_4064601.sql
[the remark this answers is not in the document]
harisang: Actually thought about this and decided to follow your suggestion and remove this unnecessary decomposition.

cowprotocol/accounting/slippage/slippage_prices_4064601.sql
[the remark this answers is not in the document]
harisang: Does this look any better now?

One could also easily move the intrinsic prices computation to a separate query but given this is something that will probably not be used anywhere, i am not sure if it make sense.

cowprotocol/accounting/slippage/slippage_prices_4064601.sql
```diff
@@ -25,14 +26,24 @@ with token_times as (
...
+),
+
+-- the Dune price feed; note that this is computed on Dune and is not part of the imported_price_feeds_raw table.
+dune_price_feed as (
     select -- noqa: ST06
         date_trunc('hour', minute) as hour, --noqa: RF04
         token_address,
         decimals,
-        avg(price) as price_unit,
-        avg(price) / pow(10, decimals) as price_atom
```
fhenneke, line 35: Is it intentional that `price_atom` is removed here?

harisang: The reason was that we can work with units and at the very end introduce the price of an atom as we know the number of decimals. This was to avoid repetition of caclulations (e.g., computing the median for the price_unit and price_atom)

cowprotocol/accounting/slippage/slippage_prices_4064601.sql
```diff
@@ -135,5 +187,5 @@ native_token_prices as (
 )
 
 select * from prices
-union all
+union distinct
```
fhenneke, line 190: Why was this change required? Before, the entries in the prices table were always distinct from entries in the native_token_prices table since the token address was different. Does this still hold?

If not, there might be additional issues due to a token having multiple prices at the same time, leading to duplicate entries when joining.

harisang: Tbh i didn't think too much about it and it just looked more correct to me (in the sense that if you provide a prices table where native tokens are included, then you should not use the corresponding native_token_prices entries). In the current state, i don't think it can ever happen that the two table have common entries so i will switch back to union all

fhenneke:
Is there a way to test this query?

harisang:
For the median part of the code, i added this "test"
https://dune.com/queries/4646457

cowprotocol/accounting/price_feed/raw_price_feed_query_4252674.sql
```diff
@@ -1,56 +1,46 @@
...
         cast(replace(time, 'T', ' ') as timestamp) as time, --noqa: RF04
-        cast(price as double) as price_unit_eth,
+        'ethereum' as blockchain,
+        token_address as contract_address,
         decimals,
-        source
-    from dune.cowprotocol.dataset_price_feed_{{blockchain}}
+        source,
+        cast(price as double) as price_unit_eth
+    from dune.cowprotocol.dataset_price_feed_ethereum
```
fhenneke, line 22: This change looks good. I.e. removing `blockchain` as parameter to the query and to only have one query for all chains. The filtering per chainis then done in a later step, same as for the normal dune prices table.

If we add price feeds for other chains, `imported_prices` will become a `union all` of the different dataset tables.

cowprotocol/accounting/price_feed/raw_price_feed_query_4252674.sql
```diff
@@ -1,56 +1,46 @@
...
-from imported_prices_per_minute_with_usd_prices group by 1, 2, 3, 4
+    ippm.minute,
+    ippm.blockchain,
+    ippm.contract_address,
+    ippm.decimals,
+    ippm.source,
+    ippm.price_unit_eth * p.price as price
+from imported_prices_per_minute as ippm
+inner join prices.usd as p on ippm.minute = p.minute
+where p.contract_address = 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2 and p.blockchain = 'ethereum'
```
fhenneke, line 45: I do not understand the implications of this. On some level, this means that the output price feed contains information different from what is uploaded and is present in our database. Should we also fetch ETH prices and store them?

Maybe we have to go with this for now. But it indicates that we are not using correct abstractions somewhere (e.g. not natural currencies, not enough data, or soomething like that).

cowprotocol/accounting/price_feed/raw_price_feed_query_4252674.sql
```diff
@@ -1,56 +1,46 @@
 -- This query parses the raw price_feed table that we sync on Dune,
--- and returns a view that is based on hourly prices, similar to
--- how the Dune price feed is used for slippage accounting
+-- and returns a view that is based on minute prices, quite similar to
+-- how the Dune price.usd looks like
```
fhenneke, line 3: This approach of being consistent + adding one column on the source sounds like the right approach.

cowprotocol/accounting/price_feed/raw_price_feed_query_4252674.sql
harisang: > If we add price feeds for other chains, imported_prices will become a union all of the different dataset tables.

yes, this is the idea here

cowprotocol/accounting/price_feed/raw_price_feed_query_4252674.sql
```diff
@@ -1,56 +1,46 @@
...
+    ippm.minute,
+    ippm.blockchain,
+    ippm.contract_address,
+    ippm.decimals,
+    ippm.source,
+    ippm.price_unit_eth * p.price as price
+from imported_prices_per_minute as ippm
+inner join prices.usd as p on ippm.minute = p.minute
+where p.contract_address = 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2 and p.blockchain = 'ethereum'
+order by ippm.minute desc --noqa: AM06
```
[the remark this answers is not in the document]
harisang: Not really, so i could just remove it as well

cowprotocol/accounting/price_feed/raw_price_feed_query_4252674.sql
fhenneke: Lets better stick with what you did here.

In general, we typically use native prices throughout the accounting and only use USD prices in some of the output intended for human consumption. But the consistency with `prices.usd` looks like a good enough reason to stick with what you wrote.

cowprotocol/accounting/slippage/slippage_prices_4064601.sql
harisang: I like the suggestion. Just applied

fhenneke:
Do you have a test query on dune for some of the final queries, e.g. raw slippage from query 4059683?

I still do not quite get what the final design would be and what the intermediate testing would look 
like. Do we just run every query twice and compare? Or have one more query which calls slippage with different price feed parameters? Will we remove the price feed parameter at some point? How does it help with fixing prices?

harisang:
So, i forked basically all queries that this PR changes as follows:
- slippage_4070065 -> https://dune.com/queries/4668024
- slippage_prices -> https://dune.com/queries/4646513
- raw_slippage -> https://dune.com/queries/4646519
- unusual_slippage -> https://dune.com/queries/4671700

There are some minor changes in some queries so that their output is more interesting for testing purposes

harisang:
> I still do not quite get what the final design would be and what the intermediate testing would look
> like. Do we just run every query twice and compare? Or have one more query which calls slippage with different price feed parameters? Will we remove the price feed parameter at some point? How does it help with fixing prices?

My goal would be to eventually use the "new price feed". What this will be is not clear yet. Currently it is the median of some price feeds; we could add more price feeds or take some other function of the existing ones, or average out over longer periods of time, etc etc. But the goal is to replace the Dune price feed eventually.

The problem itself is a bit underspecified, that's why i don't have a clear answer to "how does it help with fixing prices". 

A concrete first goal for me would be to kill outliers, i.e., not have to manually exclude txs from the accounting because slippage is extremely off.

fhenneke:
I checked the raw slippage per transaction query and there seems to be a significant difference between the old and new query with the dune price feed (e.g. for transaction `0x89aae16f7b5650fb4f4cc59fe1bb63d6c622a205a4ac8b344db748e800ce7a47` or `0x8f941c8512d281efee3e76f5d388a76af40990812478c5840b3dda538c3f3d9e`). And the differences between the two price feeds in the new query seems small (but thta would kind of be expected).
old:
image
new with dune price feed:
image
new with median price feed:
image

Do you know where those differences are coming from?

fhenneke:
I think the issue is with using prices per unit instead of prices per atom. E.g. for token `0x9de1c3d446237ab9baff74127eb4f303802a2683`, the Dune table does not contain a price nor does it contain information on decimals. The original reworked slippage accounting intentionally used prices per atom which could still be computed from exchange rates. The rework in this PR moves to prices per unit, which depends on having information on decimals. The corresponding contribution to slippage is thus ignored.

PR #66, opened 2024-11-18
Introduce new service fee tracker query

harisang:
The current service fee tracker query has proved to be problematic with the new colocated drivers. This PR proposes a new query that is meant to replace the existing one, and hopefully is much simpler to understand and debug. Also, all calculations are pushed to the Dune side, meaning that the query simply returns a true/false flag. Note that this means the solver-rewards script needs to be adapted a bit to use this query.

An important detail is that this query only looks at solvers that are part of the CoW DAO bonding pool. No other solver appears in the final list, which means that if a solver is absent from the list, it means it belongs to some other bonding pool and thus should NOT pay a service fee.

The old query can be found here: https://dune.com/queries/4017925

while the one this PR proposes can be found here: https://dune.com/queries/4298142

### DIFF
--- a/cowprotocol/accounting/rewards/mainnet/mainnet_dashboard_query_2510345.sql
+++ b/cowprotocol/accounting/rewards/mainnet/mainnet_dashboard_query_2510345.sql
@@ -133,7 +133,7 @@ service_fee_flag as (
             when service_fee then 0.85
             else 1
         end as service_fee_factor
-    from "query_4298142(time='{{end_time}}')"
+    from "query_4298142(start_time='{{start_time}}', end_time='{{end_time}}')"
 ),
 
 combined_data_after_service_fee as (


cowprotocol/accounting/rewards/mainnet/reduced_bonding_pools_query_4065709.sql
```diff
@@ -3,22 +3,67 @@
...
+        'Reduced-CoW-DAO' as pool_name,
+        from_hex('0x3075F6aab29D92F8F062A83A0318c52c16E69a60') as pool_address,
+        from_hex('0x6bf97aFe2D2C790999cDEd2a8523009eB8a0823f') as solver_address,
+        timestamp '2024-10-21 03:33:00' as creation_date
+    union distinct
+    select
+        'barn-Portus' as solver_name,
+        'Reduced-CoW-DAO' as pool_name,
+        from_hex('0x3075F6aab29D92F8F062A83A0318c52c16E69a60') as pool_address,
+        from_hex('0x5131590ca2E9D3edC182581352b289dcaE83430c') as solver_address,
```
bram-vdberg, line 51: Should we remove this since we un-whitelisted their barn address?

cowprotocol/accounting/rewards/mainnet/service_fee_query_DEPRECATED_4017925.sql
```diff
@@ -1,3 +1,7 @@
+------------ DEPRECATED!
```
fhenneke, line 1: Adding this comment is useful. Keeping the file under version control probably not.

It the goal of this to allow for a non-breaking transition?

cowprotocol/accounting/rewards/mainnet/service_fee_query_4298142.sql
```diff
@@ -0,0 +1,85 @@
+-- This query computes which solvers need to pay service fee
+-- by breaking down the calculation in small (and somewhat tedious) steps
+-- We stress the query is not efficient (and recomputes same things along the way)
+-- but we believe it is at least straightforward to check its correctness, and given
+-- that it's fast anyways, we decided to go with this.
+
+-- Parameters:
+-- {{time}}: the end date of an accounting week
```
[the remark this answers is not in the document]
harisang: I want the query to pick on solvers that join during an accounting week. E.g., a solver that is vouched for and whitelisted on a Wednesday.

cowprotocol/accounting/rewards/mainnet/reduced_bonding_pools_query_4065709.sql
harisang: I think it's fine to keep them as is. Once we unvouch these addresses, they should have no effect as the final result is a left-joining the active solvers with the reduced bonding pools

cowprotocol/accounting/rewards/mainnet/service_fee_query_DEPRECATED_4017925.sql
harisang: Yeah good point. I guess in order to add this comment, we either need to first merge this PR and then have a separate one delete the relevant file, or manually add this comment directly on Dune.

Dune doesn't allow the deletion of queries, right?

fhenneke: One can _archive_ queries. I think such queries cannot be used from within other queries.

cowprotocol/accounting/rewards/mainnet/service_fee_query_4298142.sql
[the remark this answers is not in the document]
fhenneke: I would probably add a `start_time` (for checking the service fee status) and an `end_time` (to be consistent with the strange parameter for the vouching query). That is then also consistent with all other accounting queries and does not hardcode a fixed 7 day accounting period.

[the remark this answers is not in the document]
harisang: Just pushed a commit with this change

cowprotocol/accounting/rewards/mainnet/mainnet_dashboard_query_2510345.sql
```diff
@@ -133,22 +133,22 @@ service_fee_flag as (
             when service_fee then 0.85
             else 1
         end as service_fee_factor
-    from "query_4017925"
+    from "query_4298142(time='{{end_time}}')"
```
fhenneke, line 136:
```suggestion
    from "query_4298142(start_time='{{start_time}}', end_time='{{end_time}}')"
```

fhenneke (approved):
Except for calling the service fee query from the main dashboard, this looks good to me.

PR #67, opened 2024-11-19
Update CoW AMM bounty query to handle uncatalogued solvers

harisang:
The current version of the query is removing solvers that do not have their name "registered" on Dune. This PR addresses this by replacing an "Uncatalogued" solver name with the corresponding solver address, so that the solver is still included int he final results

Query has already been updated as it was needed for today's payout.

https://dune.com/queries/4031724?cow_budget_n26d66=28650&start_time_d83555=2024-11-12+00%3A00%3A00&end_time_d83555=2024-11-19+00%3A00%3A00


harisang:
Since this is a temporary query, i wouldn't really be worried too much about duplicated code. If we decide to proceed with vouching on multiple chains, then indeed we could only work with addresses and then these problems would be gone.

PR #69, opened 2024-11-26
Add protocol cost coverage query

harisang:
This PR adds the weekly "cost coverage" query that we use in the main payouts dashboard on mainnet under version control. It also slightly changes it so that it can account for the daily fee withdrawals at 00:00 UTC that we recently introduced.

Here is the current version of the query:
https://dune.com/queries/2601653

Line 11 is the only line that is changed with this PR

### DIFF
--- a/cowprotocol/accounting/cost_coverage/cost_coverage_2601653.sql
+++ b/cowprotocol/accounting/cost_coverage/cost_coverage_2601653.sql
@@ -1,3 +1,10 @@
+-- This query computes some simple cost coverage statistics
+-- for a week. Although time range can be specified by the user
+-- this query is aligned with the weekly payouts so its results
+-- are useful when start/end date aligns with an accounting period 
+-- Parameters:
+-- {{start_time}}: the start date of an accounting week
+-- {{end_time}}: the end date of an accounting week
 with
 purchased_eth as (
     select
@@ -22,8 +29,8 @@ transferred_eth as (
         "from" = 0x9008d19f58aabd9ed0d60971565aa8510560ab41
         and to = 0xa03be496e67ec29bc62f01a428683d7f9c204930
         and contract_address = 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2
-        and evt_tx_hash not in (select tx_hash from cow_protocol_ethereum.trades where block_date between (timestamp '{{start_time}}' + interval '2' day) and (timestamp '{{end_time}}' + interval '2' day))
-        and evt_block_time between (timestamp '{{start_time}}' + interval '2' day) and (timestamp '{{end_time}}' + interval '2' day)
+        and evt_tx_hash not in (select tx_hash from cow_protocol_ethereum.trades where block_date between (timestamp '{{start_time}}' + interval '12' hour) and (timestamp '{{end_time}}' + interval '12' hour))
+        and evt_block_time between (timestamp '{{start_time}}' + interval '12' hour) and (timestamp '{{end_time}}' + interval '12' hour)
     group by evt_block_time
 ),
 
@@ -52,10 +59,6 @@ outgoing_eth as (
         and success = true
         and (lower(call_type) not in ('delegatecall', 'callcode', 'staticcall') or call_type is null)
         and block_time between (timestamp '{{start_time}}' + interval '2' day) and (timestamp '{{end_time}}' + interval '2' day)
-        -- Excluding 80 ETH transfer due to 
-        -- https://snapshot.org/#/cow.eth/proposal/0x79fdcc006030d50ab0ffe0ffd7c474a409eb70448d1c8eba58919af7559a876e
-        -- https://etherscan.io/tx/0x86f101b7ffa11c734ffe117bb2f0e4b377260f3fc3d8517e91d01126eed12980
-        and tx_hash != 0x86f101b7ffa11c734ffe117bb2f0e4b377260f3fc3d8517e91d01126eed12980
 ),
 
 outgoing_cow as (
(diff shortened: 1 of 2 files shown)


cowprotocol/accounting/cost_coverage/cost_coverage_2601653.sql
```diff
@@ -0,0 +1,117 @@
...
+purchased_eth as (
+    select
+        block_time,
+        sum(units_bought) as fees
+    from cow_protocol_ethereum.trades
+    where
+        buy_token_address in (0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2, 0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee)
+        and trader in (0x84e5c8518c248de590d5302fd7c32d2ae6b0123c, 0x9008D19f58AAbD9eD0D60971565AA8510560ab41)
+        and receiver = 0xa03be496e67ec29bc62f01a428683d7f9c204930
+        and block_date between (timestamp '{{start_time}}' + interval '12' hour) and (timestamp '{{end_time}}' + interval '12' hour)
```
fleupold, line 11: Can you comment in the code why we use 12 hours here (it may be a surprising for an unsuspecting reader).

cowprotocol/accounting/cost_coverage/cost_coverage_2601653.sql
```diff
@@ -0,0 +1,117 @@
...
+    select
+        evt_block_time as block_time,
+        sum(value / 1e18) as fees
+    from erc20_ethereum.evt_Transfer
+    where
+        "from" = 0x9008d19f58aabd9ed0d60971565aa8510560ab41
+        and to = 0xa03be496e67ec29bc62f01a428683d7f9c204930
+        and contract_address = 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2
+        and evt_tx_hash not in (select tx_hash from cow_protocol_ethereum.trades where block_date between (timestamp '{{start_time}}' + interval '2' day) and (timestamp '{{end_time}}' + interval '2' day))
+        and evt_block_time between (timestamp '{{start_time}}' + interval '2' day) and (timestamp '{{end_time}}' + interval '2' day)
```
fleupold, line 26: This also needs to change doesn't it (since we are also transferring WETH with every fee harvest)?

cowprotocol/accounting/cost_coverage/cost_coverage_2601653.sql
```diff
@@ -0,0 +1,117 @@
...
+    from ethereum.traces
+    where
+        "from" = 0xa03be496e67ec29bc62f01a428683d7f9c204930
+        and success = true
+        and (lower(call_type) not in ('delegatecall', 'callcode', 'staticcall') or call_type is null)
+        and block_time between (timestamp '{{start_time}}' + interval '2' day) and (timestamp '{{end_time}}' + interval '2' day)
+        -- Excluding 80 ETH transfer due to 
+        -- https://snapshot.org/#/cow.eth/proposal/0x79fdcc006030d50ab0ffe0ffd7c474a409eb70448d1c8eba58919af7559a876e
+        -- https://etherscan.io/tx/0x86f101b7ffa11c734ffe117bb2f0e4b377260f3fc3d8517e91d01126eed12980
+        and tx_hash != 0x86f101b7ffa11c734ffe117bb2f0e4b377260f3fc3d8517e91d01126eed12980
```
fleupold, line 58: Can we remove this?

fleupold (approved):
Small tip for when you move entire queries under version control and then do a small change. Best process imo is
1. Move query without change
2. git commit
3. Make change
4. git commit & git push

Then in the PR you can reference the second commit hash directly which only shows the change without losing any time or doing 2 PRs.

cowprotocol/accounting/cost_coverage/cost_coverage_2601653.sql
harisang: Ohh good catch, thanks!

cowprotocol/accounting/cost_coverage/cost_coverage_2601653.sql
```diff
@@ -0,0 +1,117 @@
...
+protocol_fees_collected as (
+    select sum(cast(cast(data.protocol_fee as varchar) as int256) * data.protocol_fee_native_price / pow(10, 18) - coalesce(case when data.partner_fee_recipient is not null then cast(data.partner_fee as int256) * data.protocol_fee_native_price / pow(10, 18) end, 0)) as protocol_fee_in_eth --noqa: RF01
+    from cowswap.raw_order_rewards as r
+    inner join ethereum.blocks as b on number = block_number
+    where
+        r.block_number > 19068880
+        and data.protocol_fee_native_price > 0 --noqa: RF01
+        and b.time between timestamp '{{start_time}}' and timestamp '{{end_time}}'
+        and r.order_uid not in (select cast(order_uid as varchar) from query_3639473)
+),
```
fhenneke, line 97: This table does not seem to be included in the final result. We should either add it there or remove this code.

fhenneke (approved):
Just added some smaller comments. Otherwise it looks fine.

cowprotocol/accounting/cost_coverage/cost_coverage_2601653.sql
harisang: Good catch. Removing it

harisang: I believe it does. It is used in the `results` table, and then we `select *` from `results` at the very end.

PR #70, opened 2024-11-27
Execution quality comparison

fleupold:
This query is based off of @PoloX2021 query to compare our execution quality with Paraswap on Aave tokens (cf. https://dune.com/queries/4274948?number_tokens_t6c1ea=100) but allowing to chose arbitrary aggregators/chains to compare with.

The basic idea is to compute "t0 markout" of trades with the Dune price table for both competitors and then volume weight the ratio of those markouts to get a total edge.

I noticed an issue in the original query that was simply computing `sum(volume * improvement) / sum(volume)` that it overcounts "positive improvements" and thus doesn't lead to symmetric results (ie. you may find when comparing Paraswap to CoW that Paraswap is better, and when comparing CoW to Paraswap that CoW is better).

This is addressed by using a geometric volume weighted average (taking the logarithm and then exp-ing the result). The intuition being, that if there are two trades of $100 one in which project A was 2x better than project B, and one where project B was 2x better than project A, the total comparision should yield 1 (projects are equally good).

In the old version, we would have
```
(100 * 2) + (100 * 0.5) / (100 + 100) = 1.25
```

Now we have

```
exp((100 * ln(2)) + (100 * ln(0.5)) / (100 + 100)) = 1
```

Also would kindly ask @fhenneke for feedback on if this math checks out.

### DIFF
--- a/cowprotocol/execution_quality/.sqlfluff
+++ b/cowprotocol/execution_quality/.sqlfluff
@@ -4,5 +4,5 @@ project='cow_protocol'
 project_a='cow_protocol'
 project_b='cow_protocol'
 blockchain='ethereum'
-top_n_tokens=100
+top_n_pairs=100
 
--- a/cowprotocol/execution_quality/.sqlfluff
+++ b/cowprotocol/execution_quality/.sqlfluff
@@ -1,4 +1,8 @@
 [sqlfluff:templater:jinja:context]
 start='2024-08-20 00:00:00'
 project='cow_protocol'
+project_a='cow_protocol'
+project_b='cow_protocol'
 blockchain='ethereum'
+top_n_tokens=100
+
--- a/cowprotocol/execution_quality/.sqlflull
+++ b/cowprotocol/execution_quality/.sqlflull
@@ -1,0 +1,4 @@
+[sqlfluff:templater:jinja:context]
+start='2024-08-20 00:00:00'
+project='cow_protocol'
+blockchain='ethereum'
--- a/cowprotocol/execution_quality/aggregator_markout_4334277.sql
+++ b/cowprotocol/execution_quality/aggregator_markout_4334277.sql
@@ -6,33 +6,36 @@
 --  {{start}} - Start date for when trades should be counted
 --  {{blockchain}} - The chain on which trades should be counted
 
-with trades as(
-    select 
-        block_time, 
-        amount_usd, 
-        (token_bought_amount/token_sold_amount) as project_price,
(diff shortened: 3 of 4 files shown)


PoloX2021:
LGTM

About the math, the 2 successive averages work as if they were done at the same time. It is effectively changing from arithmetic to geometic, which seems better as we are averagin ratios.

For the query comparing project a and b, we filter the top x tokens for project a, but the average is weight by the volumes for b. Is that on purpose?

fleupold:
That's a good catch! I can definitely order it by top x tokens for the same project. I took the volume logic from your original query and understand that we have to decide for one of the two volumes, right? Do you think project_a (or even an average) makes more sense here?

PoloX2021:
I think that we have to chose either one of the volumes. I am not sure that a combination of both volumes would make much sense. 
 
 An alternative could be to select the x top tokens for project a and select these same tokens for project b. Then we compute an index relative to dune prices for both dex agregators, instead of `exp(sum(project_b_volume*ln(project_a_advantage))/sum(project_b_volume))`
we compute `exp(sum(project_b_volume*ln(p2.dune_price_ratio))/sum(project_b_volume)) / exp(sum(project_a_volume*ln(p1.dune_price_ratio))/sum(project_a_volume)) `

There is no perfect solution to this, but I feel that getting an absolute performance metric for each project might be more natural than getting a performance metric relative to another project

fhenneke:
I am a bit out of the loop on this performance stuff, so feel free to ignore if it misses the mark.

Regarding the choice of arithmetic and geometric mean: Suposse there are 1000 trades on the same token pair we average over.
- The arithmatic mean makes sense to me when thinking about a case where 1000 individual users use a project. It then corresponds to the relative value gained or lost by using the project. In the case of 2 users, one might get 200 instead of 100, another might get 50 instead of 100, the total value generated is 50, i.e., 25% of the volume.
- The geometric mean makes sense to me when thinking about a case where one user does 1000 consecutive trades, using the same funds. It then corresponds to the relative value gained or lost by that user. In the case of the two trades above (100->200 and 100->50), one can think of the second trade bein representative of a trade 200->100 which leads to a sequence 100->200->100 with a relative performance of 0%.

Since one is probably somewhere between these extreme cases, there is no one correct choice. I would have tended towards the arithmetic mean.

Both metrics can be aggregated into one number per project. The relative performance can then be compared between protocols by taking quotients.

PoloX2021 (approved):
Seems really nice. Good idea using global DeFi volume instead of project's volume

PR #72, opened 2024-11-27
CoW AMM monitoring

PoloX2021:
Add CoW AMMs to the apr monitoring
Increase the number of Uni style to monitor


cowamm/kpi/competitors/cow_amm/cow_amm_kpi_4340428.sql
```diff
@@ -0,0 +1,24 @@
+-- Computes volume, tvl and APR for CoW AMM pools
+-- APR is measured as the fees earned per $ invested, over the last 24 hours, projected over 1 year
+-- Parameters:
+-- {{blockchain}}: The blockchain to query
+-- {{competitor_end_time}}: The end time of the time window (end_time - 1 day; end_time), defaults to now()
```
fleupold, line 5: I'm just realising that this value isn't really useful unless it's also populated from the query above. Otherwise you won't be able to call the data pipeline for a previous date.

cowamm/kpi/competitors/cow_amm/tvl_all_cow_amms_4340356.sql
```diff
@@ -0,0 +1,82 @@
+-- Computes the balances and their dollar value of all CoW AMM pools at every transaction where it changed
```
fleupold, line 1: If this query is fast enough, can we replace `query_4059700` with a simpler one that is calling into this one and simply adds a where clause?

cowamm/.sqlfluff
```diff
@@ -8,5 +8,5 @@ start_time='2024-08-20 00:00:00'
 end_time='2024-08-27 00:00:00'
 results=final_results_per_solver,cow_surplus_per_batch
 cow_budget=30000
-number_of_pools=500
+number_of_pools=1000
```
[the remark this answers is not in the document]
PoloX2021: Actually, in the sync jobs we are not specifying the number_of_pools, but simply using the default value. So, I thought that changing the default value would create the intended effect

cowamm/kpi/competitors/cow_amm/tvl_all_cow_amms_4340356.sql
PoloX2021: It's not very fast: 9 minutes vs <1 minute for `query_4059700`

cowamm/.sqlfluff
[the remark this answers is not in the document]
fleupold: If you change the default value in dune then this works (however, here you are just changing the hint for the sql linter)

fleupold (approved):

PR #76, opened 2024-11-29
Change prices.usd to prices.minute

fleupold:
Since `prices.usd` is deprecated (https://docs.dune.com/data-catalog/curated/prices/overview).

Also add nitpicker that will comment whenever a new query is checked in with the legacy table.


harisang (approved):
looks good! Hopefully performance will not be affected by this change

PR #83, opened 2024-12-04
Adapt vouch query to other chains

fhenneke:
This PR allows for using the vouching query on other chains.

Results can be checked with [this test query](https://dune.com/queries/4372569?end_time_d83555=2024-12-04+00%3A00%3A00&blockchain_t6c1ea=gnosis).


harisang (approved):

harisang:
Looks good!

PR #84, opened 2024-12-05
Extend service fee query to all chains

harisang:
This PR adds the service fee query and the reduced bonding pools query to work for all chains


fhenneke (approved):
Generally  looks reasonable.

I still think that hard-coding all start dates would be a better solution.

PR #89, opened 2024-12-09
Optimizing competitor kpis

PoloX2021:
Updated the queries to fit the new table prices.minute to use them more efficiently. 
For total TVL, switched to the table prices.day: less expensive and gets an average over the time frame for the study

### DIFF
--- a/cowamm/kpi/competitors/curve/curve_kpis_4232873.sql
+++ b/cowamm/kpi/competitors/curve/curve_kpis_4232873.sql
@@ -3,22 +3,16 @@
 -- Parameters:
 -- {{blockchain}}: The blockchain to query
 -- {{competitor_end_time}}: The end time of the time window (end_time - 1 day; end_time), defaults to now()
-select
-    contract_address,
-    tvl,
-    fee,
-    volume,
-    apr
-from (
+
+with accumulated_kpis as (
     select
-        contract_address,
+        r.contract_address,
         fee,
         tvl,
-        latest_per_pool,
-        sum(amount_usd) over (partition by contract_address order by latest_per_pool) as volume,
-        365 * sum(amount_usd * fee / tvl) over (partition by contract_address order by latest_per_pool) as apr
-    -- The first call to 4232976 gets the tvl after each tx to compute volume/tvl
-    from "query_4232976(blockchain='{{blockchain}}')" as r
+        sum(amount_usd) over (partition by r.contract_address order by latest_per_pool desc) as volume,
+        365 * sum(amount_usd * fee / (reserve0 * p0.price * power(10, -p0.decimals) + reserve1 * p1.price * power(10, -p1.decimals))) over (partition by r.contract_address order by latest_per_pool desc) as apr,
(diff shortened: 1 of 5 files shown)


cowamm/kpi/competitors/curve/curve_largest_2token_pools_4232976.sql
```diff
@@ -54,46 +54,42 @@ reserves as (
...
-        fee,
-        latest_per_pool,
         (reserve0 * p0.price / pow(10, p0.decimals)) + (reserve1 * p1.price / pow(10, p1.decimals)) as tvl
     from reserves as r
-    inner join prices.minute as p0
+    --using the daily value to get a better representation of the TVL over the 24 hour period
+    inner join prices.day as p0
         on
-            date_trunc('minute', block_time) = p0.timestamp
+            p0.timestamp = date_trunc('day', case when ('{{end_time}}' = '2100-01-01' or date('{{end_time}}') = date_trunc('day', now())) then now() - interval '1' day else date('{{end_time}}') end)
```
fleupold, line 78: this logic seems unnecessarily complex. Isn't this the same as 

`least(date('{{end_time}}'), now() - interval '1' day)`

fleupold: Also please document the new parameter

cowamm/kpi/competitors/curve/curve_kpis_4232873.sql
```diff
@@ -29,8 +24,28 @@ from (
...
+            and date_trunc('minute', r.evt_block_time) = p0.timestamp
+    inner join prices.minute as p1
+        on
+            r.token1 = p1.contract_address
+            and date_trunc('minute', r.evt_block_time) = p1.timestamp
     where
         -- This test avoids any possible issue with reconstructing the reserves of the pool
         tvl > 0
+        and p0.timestamp >= date_add('day', -1, (case when '{{competitor_end_time}}' = '2100-01-01' then now() else timestamp '{{competitor_end_time}}' end))
+        and p0.timestamp <= (case when '{{competitor_end_time}}' = '2100-01-01' then now() else timestamp '{{competitor_end_time}}' end)
```
fleupold, line 39: Maybe using 
```
p0.timestamp between date_add('day', -1, least('{{competitor_end_time}}', now())) and least('{{competitor_end_time}}', now())
```

is slightly easier to read? Also I'm not sure if reducing the range to `now()` is actually making the query more effective (since there are no datapoints beyond now the full table scan should complete in the same time)

fleupold (commented):
Approach looks good. However, I'd like to see if we can make the date filter a bit less ugly given that we need to add it all over the place.

cowamm/kpi/competitors/curve/curve_largest_2token_pools_4232976.sql
PoloX2021: Yes, indeed :)

PoloX2021:
I changed the parameter from end_time to start_time, as it makes more sense when rounding to a date

fleupold (approved):

PR #95, opened 2024-12-16
Add new full bonding pools

harisang:
This PR adds the new full bonding pools:
- Gnosis DAO: https://etherscan.io/tx/0xe6661cc048e7f00e1abe344d34f13bd616f8e25fb137279f079d10bb0fb555ed
- Project Blanc: https://etherscan.io/tx/0x20735c4b9f12fdd110e5d4f69c039ffce06f2bd4da7c92852263f5bc13d02573

It also renames the `initial_funder` column to `creator`, as it is more accurate (creator is what we are interested in as this is the address that can vouch, while there can be multiple funders)

### DIFF
--- a/cowprotocol/accounting/rewards/mainnet/full_bonding_pools_query_4056263.sql
+++ b/cowprotocol/accounting/rewards/mainnet/full_bonding_pools_query_4056263.sql
@@ -3,19 +3,29 @@
 with
 full_bonding_pools as (
     select
-        from_hex('0x8353713b6D2F728Ed763a04B886B16aAD2b16eBD') as pool_address,
+        from_hex('0x8353713b6D2F728Ed763a04B886B16aAD2b16eBD') as pool_address, -- deprecated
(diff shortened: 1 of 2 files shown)


cowprotocol/accounting/rewards/mainnet/full_bonding_pools_query_4056263.sql
```diff
@@ -3,19 +3,29 @@
 with
 full_bonding_pools as (
     select
-        from_hex('0x8353713b6D2F728Ed763a04B886B16aAD2b16eBD') as pool_address,
+        from_hex('0x8353713b6D2F728Ed763a04B886B16aAD2b16eBD') as pool_address, -- deprecated
```
fleupold, line 6: Should we rename this to deprecated? Having Gnosis vs Gnosis DAO seems a bit awkwardly implicit

fleupold (approved):
Good idea with the rename of the column!

PR #98, opened 2024-12-20
Update and extend unusual slippage query to all chains

harisang:
This PR updates the outdated version of the query that was in this repo (the phalcon explorer view link was missing) and also extends it to work for all chains.

Changes have already been applied and can be tested here:
https://dune.com/queries/2332678?start_time_d83555=2024-12-10+00%3A00%3A00&end_time_d83555=2024-12-17+00%3A00%3A00

### DIFF
--- a/cowprotocol/accounting/rewards/mainnet/unusual_slippage_query_2332678.sql
+++ b/cowprotocol/accounting/rewards/mainnet/unusual_slippage_query_2332678.sql
@@ -7,7 +7,15 @@
 --  {{significant_slippage_value}} -- the absolute threshold above which a tx is always flagged as high-slippage
 with
 results_per_tx as (
-    select * from "query_4070065(blockchain='ethereum',start_time='{{start_time}}',end_time='{{end_time}}',slippage_table_name='slippage_per_transaction')"
+    select * from "query_4070065(blockchain='{{blockchain}}',start_time='{{start_time}}',end_time='{{end_time}}',slippage_table_name='slippage_per_transaction')"
+),
+
+url_helper as (
+    select
+        case
+            when '{{blockchain}}' = 'ethereum' then 'eth'
+            else '{{blockchain}}'
+        end
 )
 
 select  --noqa: ST06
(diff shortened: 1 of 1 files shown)


cowprotocol/accounting/rewards/mainnet/unusual_slippage_query_2332678.sql
```diff
@@ -7,27 +7,43 @@
...
         '&start_time={{start_time}}',
         '&end_time={{end_time}}',
         '&slippage_table_name=raw_slippage_breakdown',
         '" target="_blank">link</a>'
     ) as token_breakdown,
-    rpt.tx_hash,
+    concat(
+        '<a href="https://phalcon.blocksec.com/explorer/tx/',
+        (select * from url_helper),
+        '/',
```
fleupold, line 35:
```suggestion
        '/0x',
```

cowprotocol/accounting/rewards/mainnet/unusual_slippage_query_2332678.sql
```diff
@@ -7,27 +7,43 @@
...
+        to_hex(rpt.tx_hash),
+        '" target="_blank">',
```
fleupold, line 37: Caught my eye that the `to_hex` makes us lose the `0x` prefix. Seems to work regardless but this is cleaner.

```suggestion
        '" target="_blank">0x',
```

fleupold (approved):
Nice, would it be possible to add a column with "network + protocol fee" to this table? This would make it easier to review as from the phalcon link I get the total slippage (including fee) from which I have to deduct the fee in order to check if the value align with our slippage accounting.

harisang:
> Nice, would it be possible to add a column with "network + protocol fee" to this table? This would make it easier to review as from the phalcon link I get the total slippage (including fee) from which I have to deduct the fee in order to check if the value align with our slippage accounting.

Will address it in a follow-up PR. Opened a relevant issue
https://github.com/cowprotocol/dune-queries/issues/101

PR #100, opened 2024-12-22
Add Base WETH withdrawal and deposit events tracking for Balance Changes query

harisang:
This PR extends query 4021257 for WETH withdrawals and deposits on Base.


bram-vdberg (approved):
LGTM

PR #102, opened 2024-12-23
Extend main rewards query to all chains

harisang:
This PR extends the [main query](https://dune.com/queries/2510345/4131517) that is used to verify the rewards to all chains.

It also slightly restructures a subfolder


cowprotocol/accounting/rewards/main_rewards_dashboard_query_2510345.sql
```diff
@@ -1,29 +1,33 @@
...
 batch_rewards as (
-    select  --noqa: ST06
-        from_hex(rbr.solver) as winning_solver,
-        -- Unpacking the needed data
-        cast(cast(rbr.data.fee as varchar) as int256) as fee,  -- noqa: RF01
-        cast(cast(rbr.data.execution_cost as varchar) as int256) as execution_cost,  -- noqa: RF01
-        cast(cast(rbr.data.capped_payment as varchar) as int256) as capped_payment  -- noqa: RF01
-    from cowswap.raw_batch_rewards as rbr
+    select
+        rbr.solver as winning_solver,
```
bram-vdberg, line 21: Why don't we need to use the `from_hex()` and `cast()` methods anymore?

bram-vdberg (approved):
LGTM!

cowprotocol/accounting/rewards/main_rewards_dashboard_query_2510345.sql
harisang: For the solver, the reason we were using from_hex() was that the `solver` was stored as a string (and not as varbinary) in the cowswap.raw_batch_rewards table.

As for the different casts for fees, executions costs etc, i think they were just unnecessary, and an overkill. It looks like we wanted to get to an int256 but the queries also work with the decimal(38,0) type that is what we use here anyways.

Definitely this is not very well-thought through, so we could also make the raw order/batch tables return int256 instead.

PR #103, opened 2024-12-23
Add Base WETH to slippage prices

harisang:
This PR adds the WETH Base address to the slippage prices query. Change has been already applied:
https://dune.com/queries/4064601


fleupold (approved):
Does this lead to an SQL error when the case is missing? Otherwise, we should probably add a default case that throws a descriptive error so that it's easy to find this missing spot.

PR #105, opened 2024-12-30
Extend cost coverage query to all networks

harisang:
This PR updates the cost coverage query that shows up in the main payouts dashboard. 

Query has been updated and can be tested here: https://dune.com/queries/2601653?start_time_d83555=2024-12-31+00%3A00%3A00&end_time_d83555=2025-01-07+00%3A00%3A00

~I still want to do some sanity checks about what exactly we are computing at the very end; in case you have any comments/feedback, let me know~

### DIFF
--- a/cowprotocol/accounting/cost_coverage/cost_coverage_2601653.sql
+++ b/cowprotocol/accounting/cost_coverage/cost_coverage_2601653.sql
@@ -3,18 +3,52 @@
 -- this query is aligned with the weekly payouts so its results
 -- are useful when start/end date aligns with an accounting period 
 -- Parameters:
--- {{start_time}}: the start date of an accounting week
--- {{end_time}}: the end date of an accounting week
+--  {{start_time}}: the start date of an accounting week
+--  {{end_time}}: the end date of an accounting week
+--  {{blockchain}}: network to run the analysis on
 with
-purchased_eth as (
+wrapped_native_token as (
+    select
+        case '{{blockchain}}'
+            when 'ethereum' then 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2 -- WETH
+            when 'gnosis' then 0xe91d153e0b41518a2ce8dd3d7944fa863463a97d -- WXDAI
+            when 'arbitrum' then 0x82af49447d8a07e3bd95bd0d56f35241523fbab1 -- WETH
+            when 'base' then 0x4200000000000000000000000000000000000006 -- WETH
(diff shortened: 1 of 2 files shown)


cowprotocol/accounting/cost_coverage/cost_coverage_2601653.sql
```diff
@@ -78,46 +112,44 @@ conversion_prices as (
...
 results as (
     select
-        (select eth from outgoing_eth) as outgoing_eth,
+        (select native_token from outgoing_native_token) as outgoing_native_token,
         (select cow from outgoing_cow) as outgoing_cow,
-        (select eth from outgoing_eth) + (select cow from outgoing_cow) * (select cow_price / eth_price from conversion_prices) as total_outgoing_eth,
-        (select fees from fees_eth) as fees_eth,
-        (select latest_withdrawal_time from fees_eth) as last_withdrawal,
-        (select protocol_fee_in_eth from protocol_fees_collected) as protocol_fee_in_eth
+        (select native_token from outgoing_native_token) + (select cow from outgoing_cow) * (select cow_price / native_token_price from conversion_prices) as total_outgoing_native_token,
```
fhenneke, line 141: This column seems to be null on arbitrum and base. Potentially an issue with conversion prices?

harisang: Good point. Indeed, the COW price is missing from the prices.usd table for Arbitrum and Base.

Tbh, this is not used anywhere at the moment for the verification of the payouts, so we could still proceed as is.

fleupold:
> Do you think we should include the CoW Swap UI fee in the protocol_fee amount this query reports?

Probably not as it's strictly speaking not a protocol fee. I think it's fine to only show it in the partner fee section (where we also show the proceeds from the partner fee share). Let's document this though, as it can be easy to miss.

cowprotocol/accounting/cost_coverage/cost_coverage_2601653.sql
harisang: One option would be to do what we are doing in the rewards calculations, which is to only consider the dollar value of COW on ethereum.
https://github.com/cowprotocol/dune-queries/blob/6ee39f14eb74fee770fe9b79ceef7b18fc5e5dba/cowprotocol/accounting/rewards/main_rewards_dashboard_query_2510345.sql#L57

fhenneke (approved):
LGTM

PR #108, opened 2025-01-03
[Revenue] Per Chain Queries

fleupold:
Adds three important revenue queries per chain
1. Every trade with its potential fee parts
2. Daily totals per part
3. All time totals per part

Those queries will be used to unify the revenue dashboard and show consistent stats for all chains.


bram-vdberg (approved):
Not sure if I'm the best person to review this, but it looks good to me.

PR #114, opened 2025-01-07
Fix parameter name in rewards dashboard query

harisang:
This PR renames some parameter in the main rewards query


fhenneke (approved):
LGTM

Is the dashboard called from somewhere automatically? Those places would also need to rename the quote parameter.

harisang:
> LGTM
> 
> Is the dashboard called from somewhere automatically? Those places would also need to rename the quote parameter.

Yes, we need to fix this particular query refresh on the solver rewards' side, which is done here:
https://github.com/cowprotocol/solver-rewards/blob/2bbd634d1a6f021ae95c2c74550c09ca09f7c17b/src/fetch/dune.py#L112

PR #120, opened 2025-01-16
Lp feed for prices

PoloX2021:
The goal of this PR is to get all the Dune data that would be necessary to compute all the performance metrics for CoW AMMs.

3 Dune queries:
- To get all the info for CoW AMMs
- To get all the different Uni Style pools addresses
- To get all the info for Uni Style AMMs

2 sql queries to merge this data with the table previously stored in our database

1 that just lays out the data in the final table


cowamm/profitability/lp_prices/cow_amm_lp_4420687.sql
```diff
@@ -0,0 +1,105 @@
...
+with date_range as (
+    select t.day
+    from
+        unnest(sequence(
+            date(timestamp '{{start}}'),
+            date(timestamp '{{end}}')
+        )) t (day) --noqa: AL01
+),
+
+--use the materialized view of query_3959044
```
fleupold, line 22:
```suggestion
query_3959044 needs to be materialized due to [... explanation of performance considerations]
```

cowamm/profitability/lp_prices/cow_amm_lp_4420687.sql
```diff
@@ -0,0 +1,105 @@
...
+cow_amm_pool as (
+    select
+        created_at,
+        token_1_address as token0,
+        token_2_address as token1,
+        address as contract_address
+    from query_3959044
```
fleupold, line 29: Make it explicit that the materialized version needs to be used.

```suggestion
    from dune.cowprotocol.result_balancer_co_w_am_ms
```

cowamm/profitability/lp_prices/cow_amm_lp_4420687.sql
```diff
@@ -0,0 +1,105 @@
+-- Issues with the other version (4420687)
```
fleupold, line 1: What is the other version? This query references the one we are using here? Is this comment cleaned up?

Also there is no high level description of what this query does at the top, which we should do for all queries.

cowamm/profitability/lp_prices/uni_style_lp_4420675.sql
```diff
@@ -0,0 +1,91 @@
+--Parameters
```
fleupold, line 1: summary of what the query does

cowamm/profitability/lp_prices/uni_style_pools_4420646.sql
```diff
@@ -0,0 +1,44 @@
+-- Finds all the Uniswap style pool: Uniswap, Sushiswap, Pancakeswap
```
fleupold, line 1: This logic is already part of https://github.com/cowprotocol/dune-queries/blob/6ee39f14eb74fee770fe9b79ceef7b18fc5e5dba/cowamm/kpi/competitors/uni_swap_style/largest_uni_style_pools_4303563.sql#L8

Let's not repeat ourselves, but make this query also call this one to only have all the uni clone factory contract addresses defined in one place.

cowamm/profitability/lp_prices/uni_style_lp_4420675.sql
```diff
@@ -0,0 +1,91 @@
...
+    token1,
+    d.day,
+    price0.price as price0,
+    price1.price as price1,
+    price0.decimals as decimals0,
+    price1.decimals as decimals1,
+    coalesce(l.lp_transfer, 0) as lp_transfer,
+    --fill in the days without reserves
+    last_value(s.reserve0) ignore nulls over (partition by p.contract_address order by d.day range between unbounded preceding and current row) as reserve0,
+    last_value(s.reserve1) ignore nulls over (partition by p.contract_address order by d.day range between unbounded preceding and current row) as reserve1
```
fleupold, line 64: fot cow amms you are tracking transfer0/1 instead of reserves. Wouldn't it be better for consistency to use the same?

PoloX2021: Apparently it slows a lot the query to do so. Even when accessing only once the evt_transfer, it fails for small periods of time.

cowamm/profitability/lp_prices/cow_amm_lp_4420687.sql
```diff
@@ -0,0 +1,104 @@
...
+left join prices.day as price0
+    on
+        d.day = price0.timestamp
+        and p.token0 = price0.contract_address
+left join prices.day as price1
+    on
+        d.day = price1.timestamp
+        and p.token1 = price1.contract_address
+where
+    coalesce(price0.blockchain, '{{blockchain}}') = '{{blockchain}}'
```
harisang, line 103: is it possible that the blockchain column of the `prices.day` table is empty for some rows?

cowamm/profitability/lp_prices/cow_amm_lp_4420687.sql
```diff
@@ -0,0 +1,104 @@
...
+lp_balance_delta as (
+    select
+        contract_address,
+        date(evt_block_time) as "day",
+        sum(case when "from" = 0x0000000000000000000000000000000000000000 then value else -value end) as lp_transfer
+    from erc20_{{blockchain}}.evt_transfer
+    where
+        contract_address in (select contract_address from cow_amm_pool)
+        and ("from" = 0x0000000000000000000000000000000000000000 or "to" = 0x0000000000000000000000000000000000000000)
+        and evt_block_time >= date(timestamp '{{start}}')
```
harisang, line 41: Not sure if this would really be an optimization or not, but would it speed things up if you first define a `block_range` table and then define the intervals by referencing these blocks instead of doing these date conversions? I suspect the gain would be small, if any, so this is just a thought

cowamm/profitability/lp_prices/uni_style_lp_4420675.sql
```diff
@@ -0,0 +1,93 @@
+-- This query gets the reserve of every Uni style pool at the end of a day
```
harisang, line 1:
```suggestion
-- This query gets the reserve of every Univ2 style pool at the end of a day 
```

cowamm/profitability/lp_prices/uni_style_lp_4420675.sql
```diff
@@ -0,0 +1,93 @@
...
+left join prices.day as price0
+    on
+        d.day = price0.timestamp
+        and p.token0 = price0.contract_address
+left join prices.day as price1
+    on
+        d.day = price1.timestamp
+        and p.token1 = price1.contract_address
+where
+    (price0.blockchain = '{{blockchain}}' or price0.blockchain is null)
```
harisang, line 91: do you know what it means for the blockchain entry to be zero?

cowamm/profitability/lp_prices/uni_style_pools_4420646.sql
```diff
@@ -0,0 +1,44 @@
...
+                0x1097053Fd2ea711dad45caCcc45EfF7548fCB362,
+                0x02a84c1b3bbd7401a5f7fa98a384ebc70bb5749e
+            ) then 'pancakeswap'
+        when
+            contract_address in (
+                0x5c69bee701ef814a2b6a3edd4b1652cb9cc5aa6f,
+                0xf1D7CC64Fb4452F05c498126312eBE29f30Fbcf9,
+                0x8909Dc15e40173Ff4699343b6eB8132c65e18eC6
+            )
+            then 'uniswap'
```
harisang, line 22: should we be more specific, i.e, uniswapv2?

cowamm/profitability/lp_prices/uni_style_pools_4420646.sql
```diff
@@ -0,0 +1,44 @@
...
+            )
+            then 'sushiswap'
+    end as project
+from {{blockchain}}.logs
+where
+    topic0 = 0x0d3648bd0f6ba80134a33ba9275ac585d9d315f0ad8355cddefde31afa28d0e9
+    and contract_address in
+    (
+        0x1097053Fd2ea711dad45caCcc45EfF7548fCB362, --eth, pancake
+        0x02a84c1b3bbd7401a5f7fa98a384ebc70bb5749e, --arb/bas, pancake
```
harisang, line 37:
```suggestion
        0x02a84c1b3bbd7401a5f7fa98a384ebc70bb5749e, --arb/base, pancake
```

cowamm/profitability/lp_prices/uni_style_pools_4420646.sql
```diff
@@ -0,0 +1,44 @@
...
+        0x5c69bee701ef814a2b6a3edd4b1652cb9cc5aa6f, --eth, uni
+        0xf1D7CC64Fb4452F05c498126312eBE29f30Fbcf9, --arb, uni
+        0x8909Dc15e40173Ff4699343b6eB8132c65e18eC6, --bas, uni
```
harisang, line 40:
```suggestion
        0x8909Dc15e40173Ff4699343b6eB8132c65e18eC6, --base, uni
```

cowamm/profitability/lp_prices/uni_style_pools_4420646.sql
```diff
@@ -0,0 +1,44 @@
...
+        0xC0AEe478e3658e2610c5F7A4A2E1777cE9e4f2Ac, --eth, sushi
+        0xc35DADB65012eC5796536bD9864eD8773aBc74C4, --arb/gno, sushi
```
harisang, line 42:
```suggestion
        0xc35DADB65012eC5796536bD9864eD8773aBc74C4, --arb/gnosis, sushi
```

cowamm/profitability/lp_prices/uni_style_pools_4420646.sql
```diff
@@ -0,0 +1,44 @@
...
+        0x71524B4f93c58fcbF659783284E38825f0622859 --bas, sushi
```
harisang, line 43:
```suggestion
        0x71524B4f93c58fcbF659783284E38825f0622859 --base, sushi
```

harisang (approved):
Left some minor comments. Looks good overall, so approving, expecting comments will be addressed

cowamm/profitability/lp_prices/cow_amm_lp_4420687.sql
PoloX2021: Not 100% sure it could happen. But, for very illiquid tokens, this could be a possibility if it does not trade for a day. 
This test avoids removing the row if there is no price, which would mess up the reserves

cowamm/profitability/lp_prices/uni_style_lp_4420675.sql
PoloX2021: This test is important, because the where happens after the left join. So if there isn't a price for a token (e.g. no trades), then it filters out the line for the day.
This might be more of a precaution because of the issues with the new dune prices tables.

cowamm/profitability/lp_prices/uni_style_pools_4420646.sql
PoloX2021: Yes, will do this in a different PR

cowamm/profitability/lp_prices/cow_amm_lp_4420687.sql
PoloX2021: Apparently, the `logs` tables are partitioned by block_time and not block_number, so this should be the most efficient method.

PR #122, opened 2025-01-21
Change prices from minute to day to avoid issues in the db

PoloX2021:
The prices.minute table is completely messed up currently. 
This PR switch to prices.day. This table is less precise because we are looking at individual transactions. But it does not impact daily performance metrics.


harisang:
Out of curiosity, did you try to slice the `prices.minute `table by blockchain for example, i.e., or even by year etc? Just curious. Approving anyways as the change makes sense

harisang (approved):

PR #123, opened 2025-01-24
Fix performance kpi competitors

PoloX2021:
Improving the speed of some queries and reverting to the prices.usd table for accurate information.
This means we won't be supporting new tokens, but it's better than having false data for all tokens.

### DIFF
--- a/cowamm/kpi/competitors/cow_amm/tvl_all_cow_amms_4340356.sql
+++ b/cowamm/kpi/competitors/cow_amm/tvl_all_cow_amms_4340356.sql
@@ -2,13 +2,14 @@
 -- Parameters
 --  {{blockchain}} - chain for which the query is running
 
+--Use materialized view to avoid calling too many times the logs table, slowing the query
 with cow_amm_pool as (
     select
         created_at,
         address,
         token_1_address as token0,
         token_2_address as token1
-    from query_3959044
+    from dune.cowprotocol.result_balancer_co_w_am_ms
     where blockchain = '{{blockchain}}'
     order by 1 desc
 ),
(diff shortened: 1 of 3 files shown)


PoloX2021:
Yes, I am sure. Prices.minute is not working at the moment.

cowamm/kpi/competitors/cow_amm/tvl_all_cow_amms_4340356.sql
```diff
@@ -2,13 +2,14 @@
...
 
+--Use materialized view to avoid calling too many times the logs table, slowing the query
 with cow_amm_pool as (
     select
         created_at,
         address,
         token_1_address as token0,
         token_2_address as token1
-    from query_3959044
+    from dune.cowprotocol.result_balancer_co_w_am_ms
```
fhenneke, line 12: 😑

fhenneke:
> Prices.minute is not working at the moment.

What do you mean by that? Are values wrong? Is it too slow?

PoloX2021:
> > Prices.minute is not working at the moment.
> 
> What do you mean by that? Are values wrong? Is it too slow?

Both ;(
Although Dune advertises these tables on their website. The values are wrong, missing, there are duplicates. Apparently they are still working on them, but no date announced.

cowamm/kpi/competitors/cow_amm/tvl_all_cow_amms_4340356.sql
PoloX2021: Not sure I understand the comment. 
Is the issue witht the name of the materialized view?
Otherwise, this table does not change a lot and takes time to refresh, so it simplifies queries to have it as a materialized view.

fhenneke: Sorry for being a bit terse here. The name of that table did seem to have a lot of `_`s.

fhenneke (approved):
Changes look good to me.

PR #127, opened 2025-01-28
[Easy] All Uniswap pools

PoloX2021:
Use a base query to reference all the Uniswap pools. 
Easier to store all the parameters in the same place.
Test query [here](https://dune.com/queries/4634430).


fhenneke (approved):
Sounds good to remove as much code duplication as possible.

PR #128, opened 2025-01-29
Fix bug with removed duplicates in balance changes

fhenneke:
This PR changes how duplicates are handled in balance changes. The new query can be tested [here](https://dune.com/queries/4636979?end_time_d83555=2025-01-28+00%3A00%3A00&start_time_d83555=2025-01-21+00%3A00%3A00).

Before, with a `union distinct` all duplicate rows were removed. This led to wrongfully removing balance changes in auctions where there are multiple withdrawals of the same amount (e.g. [here](https://etherscan.io/tx/0x08afb56d779bc2034bfcc78a1d470cc82c8c55c4394fda3875d426de869f8ca2) and [here](https://etherscan.io/tx/0x74e730407d3061e60527629123c1e7b43e3e73b4ba4232bd077c7109ac5e8325)). Using `union all` removes this problem.

Comparing the new and old query for the accounting week 2025-01-21 to 2025-01-28, there are just 6 additional rows which can be traced to two settlements with duplicate withdrawals ([Dune query](https://dune.com/queries/4637074)).


harisang (approved):
Makes sense!

PR #132, opened 2025-02-04
[MEVBlocker] Prepare fee bump

fleupold:
MEVBlocker fee is increasing to 30% as of February. In order to not disrupt the billing transaction tomorrow (which should still charge 20% fee for the previous week), this PR also adjusts the "payment due " query downward to account for the old fee.

However, the per block fee that will be announced in the billing transaction and is valid for the following week already uses the 30% fee tier.

Towards the end of the week we then need to revert the change to payment due query to make sure next week the 30% fee is applied.

### DIFF
--- a/mevblocker/fees/payment_due_4005800.sql
+++ b/mevblocker/fees/payment_due_4005800.sql
@@ -15,7 +15,8 @@ WITH block_range AS (
 ),
 
 final_fee AS (
-    SELECT avg_block_fee_wei
+    --TODO: remove the 2/3 once the fee increment should also be reflected in the payment due query (not just in announcing the new fee)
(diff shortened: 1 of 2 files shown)


PoloX2021:
Pretty straightforward. LGTM
I'm not sure why it's better to comput 2/3*0.3, instead of waiting and then simply switching from 0.2 to 0.3, but seems the same

fleupold:
> I'm not sure why it's better to comput 2/3*0.3, instead of waiting and then simply switching from 0.2 to 0.3, but seems the same

The issue is that the final fee query (4002039) itself is used to compute the payment due. So we need it
1) to show the new 30% fee we advertise to builders going forward
2) also have it return the old fee for tomorrows bill computation

The easiest way I found is to manually adjust the bill downward for last week.

PR #133, opened 2025-02-08
[MEVBlocker] cleanup after fee increase to 30%

fleupold:
This PR is a follow up to #132, it 
1. Removes the temporary fee reduction (as last week we were still charging a 20% fee) for billing purposes so that as of next week we not only announce but also charge the new fee
2. Includes Titan's table for which transaction hashes were not exclusive to MEV Blocker. Those should be excluded for billing purposes.

Note, that this PR needs to be merged before next Wednesday to take effect in the payouts.

## Summary by CodeRabbit

- **Bug Fixes**  
	- Adjusted fee calculations so fees now reflect the complete average value, ensuring weekly fees and amounts due are accurate.  
	- Refined the processing of transaction data to more precisely distinguish exclusive transactions, improving the accuracy of reported results.  
	- Updated filtering logic for user transactions to include a broader scope of transactions in the analysis.  
	- Standardized data type usage across the script for consistency.  
	- Renamed variables for clarity without altering functionality in calculations.

### DIFF
--- a/mevblocker/fees/payment_due_4005800.sql
+++ b/mevblocker/fees/payment_due_4005800.sql
@@ -15,8 +15,7 @@ WITH block_range AS (
 ),
 
 final_fee AS (
-    --TODO: remove the 2/3 once the fee increment should also be reflected in the payment due query (not just in announcing the new fee)
-    SELECT avg_block_fee_wei * 2 / 3 AS avg_block_fee_wei
+    SELECT avg_block_fee_wei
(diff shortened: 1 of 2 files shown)


PoloX2021:
The query looks good.
Should we stick to capital letters or not? The linting commits seem to be doing both.

alex-zadorozhnyi:
Looks good to me

alex-zadorozhnyi (approved):

PR #135, opened 2025-02-10
Update partner fee cut

harisang:
This PR updates the partner fee cut, given the recent changes implied by [CIP-61](https://snapshot.box/#/s:cow.eth/proposal/0xbafee087f3a1d5757e44a72fa76e8624ce212623153816a936a8888741485ef5)

## Summary by CodeRabbit

- **Refactor**
	- Updated fee calculation logic to apply refined partner eligibility conditions.
	- Revised fee allocation now results in a more balanced distribution for partner fee components.


harisang:
> LGTM, would it be worth adding the partner_recipient names as a comment for clarity?

I think we discussed it is better to not explicitly label each recipient

PR #138, opened 2025-02-12
Add revenue per integration partner query

harisang:
Current query: https://dune.com/queries/3938636
is not under version control and is using the `cowswap.raw_order_rewards` table which has been deprecated.

UPDATE: Here is a query generated by chatgpt, with some mild edits, that compares the new with the old version. Results up till November 2024 should be identical modulo very small rounding errors.

https://dune.com/queries/4717753?aggregate_by_t6c1ea=month&category=third_party_data


bram-vdberg (approved):
LGTM

PR #140, opened 2025-02-18
Fix service fee query

harisang:
The query assumed, implicitly (this was not really intended), that every colocated solver with a reduced pool had initially joined the CoW DAO bonding pool at least 3 months before creating the reduced pool. This caused some issues, with Portus, for example. This is a bug, and this PR adds an explicit check so as to ensure that solvers that joined the CoW DAO bonding pool AND created a subpool within an interval of 3 months will start paying a service fee after 3 months. Specifically for Portus, there is no ambiguity around this as they joined a long time ago and only got renamed (which revealed the issue in the first place)

### DIFF
--- a/cowprotocol/accounting/rewards/service_fee_query_4298142.sql
+++ b/cowprotocol/accounting/rewards/service_fee_query_4298142.sql
@@ -121,6 +121,7 @@ select
     coalesce(e.pool_name, d.pool_name) as pool_name,
     case
         when e.creation_date > date_add('month', -3, cast('{{start_time}}' as timestamp)) then false
+        when e.creation_date <= date_add('month', -3, cast('{{start_time}}' as timestamp)) then true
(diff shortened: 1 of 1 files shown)


bram-vdberg (approved):
LGTM

fhenneke (approved):
The change looks fine.

The previous behavior makes more sense to me. It only does not work due to using names for determining joining dates. This seems like one more reason to just hard-code values instead of maintaining that query.

harisang:
> The change looks fine.
> 
> The previous behavior makes more sense to me. It only does not work due to using names for determining joining dates. This seems like one more reason to just hard-code values instead of maintaining that query.

Actually you are right. The previous version is the most reasonable in terms of interpreting the CIP and this "fix" just works because of the given situation of the existing solvers. Will rethink a bit more about this

PR #141, opened 2025-02-27
[MEV Blocker] Prepare Fee Bump

alex-zadorozhnyi:
Increase fees from 30% to 40%. At the same time keep payments due at 30% fee.
Price reduction to be reverted after next payment (5 Mar 2025).

### DIFF
--- a/mevblocker/fees/payment_due_4005800.sql
+++ b/mevblocker/fees/payment_due_4005800.sql
@@ -15,6 +15,7 @@ WITH block_range AS (
 ),
 
 final_fee AS (
+    -- TODO: remove between  2025/03/06 and 2025/03/11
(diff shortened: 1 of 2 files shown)


mevblocker/fees/payment_due_4005800.sql
```diff
@@ -15,7 +15,7 @@ WITH block_range AS (
 ),
 
 final_fee AS (
-    SELECT avg_block_fee_wei
+    SELECT avg_block_fee_wei * 3 / 4 AS avg_block_fee_wei
```
fleupold, line 18: nit 

```suggestion
    -- TODO: remove between  2025/03/06 and 2025/03/11
    SELECT avg_block_fee_wei * 3 / 4 AS avg_block_fee_wei
```

fleupold (approved):
LGTM

PR #142, opened 2025-02-28
AMMs LP history

PoloX2021:
The goal of the PR is to create one materialized view which stores the historical state of CoW AMMs and competitors.
This will speed up the queries on each pool.

This first version only gets the data for CoW AMM, uniswap, pancakeswap and sushiswap. It'll be possible to add other pools later on, using the same methodology and aggregating it in the final materialized view (e.g. Balancer & Curve).

This new version also adds the weights of the pools


harisang:
@PoloX2021 Could you fix the linting issues?

cowamm/profitability/lp_prices/amm_lp_table_4781568.sql
```diff
@@ -0,0 +1,79 @@
+--Groups all the data in a same tabble to then materialize it in a view
```
harisang, line 1:
```suggestion
--Groups all the data in one table to then materialize it in a view
```

cowamm/balancer_cow_amms_3959044.sql
```diff
@@ -3,7 +3,7 @@
 --
 -- the final table has columns
 -- - created_at: the creation timestamp
--- - blockchain: 'ethereum' or 'gnosis'
+-- - blockchain: 'ethereum' or 'gnosis' or 'arbitrum' or 'base'
 -- - address: address of Balancer CoW AMM
 -- - token_1_address: address of token with smaller address
```
harisang, line 8: As i noticed in other queries, is the standard way to reference those to use index 0-1 instead of 1-2? If so, should we switch here as well?

cowamm/profitability/lp_prices/cow_amm_lp_4420687.sql
```diff
@@ -1,17 +1,18 @@
 -- This query gets all the transfers during a day for all existing CoW AMMs:
--- Reserve tokens and lp tokens
+-- Reserve tokens and lp tokens and the beginning of every day
```
harisang, line 2: Do you mean "**at** the beginning of every day"?

cowamm/balancer_cow_amms_3959044.sql
PoloX2021: Yeah I had a doubt about this one as well... I think I've seen both conventions, but probably in dune queries, 1&2 is more used

PoloX2021: Actually, 0 & 1 seems better, and FLeupold agrees

cowamm/balancer_cow_amms_per_chain_4814818.sql
```diff
@@ -0,0 +1,72 @@
+-- This query gets all the cowamms created after the Balancer AMM launch on July 29, 2024 for a specific chain
```
harisang, line 1:
```suggestion
-- This query gets all the cowamms created after the Balancer CoW AMM launch on July 29, 2024 for a specific chain
```

cowamm/balancer_cow_amms_per_chain_4814818.sql
```diff
@@ -0,0 +1,72 @@
...
+--Parameters
+--  {{blockchain}}: The blockchain to query
+-- Hard coded start for the events scan to '2024-07-29', month of the Balancer AMMs launch
```
harisang, line 4:
```suggestion
-- Hard-coded start for the events scan to '2024-07-29', month of the Balancer AMMs launch
```

cowamm/balancer_cow_amms_per_chain_4814818.sql
```diff
@@ -0,0 +1,72 @@
...
+
+with cowamm_creations as (
+    select varbinary_substring(topic1, 1 + 12, 20) as address
+    from {{blockchain}}.logs
+    where (
+        contract_address in ( --the factory contracts
```
harisang, line 10: A slightly cleaner way of doing this would be to define a factory_contracts table where each row has an address and blockchain column. But i guess this is a bit of an overkill.

cowamm/balancer_cow_amms_per_chain_4814818.sql
```diff
@@ -0,0 +1,72 @@
...
+            0x03362f847B4fAbC12e1Ce98b6b59F94401E4588e --base
+        )
+        and topic0 = 0x0d03834d0d86c7f57e877af40e26f176dc31bd637535d4ba153d1ac9de88a7ea
+    )
+    and block_time >= cast('2024-07-29 00:00:00' as timestamp)
+),
+
+
+-- Bind events for tokens
+cowamm_weights as (
```
harisang, line 25: The name is a bit misleading (?)

cowamm/balancer_cow_amms_per_chain_4814818.sql
```diff
@@ -0,0 +1,72 @@
...
+),
+
+cowamms as (
+    select
+        blockchain,
+        address,
+        created_at,
+        token_0_address,
+        token_1_address,
+        100 * w1.weight / (w1.weight + w2.weight) as token_0_weight,
```
harisang, line 59: can we stick to the 0-1 indexing for the whole query?

cowamm/profitability/lp_prices/amm_lp_table_4781568.sql
```diff
@@ -0,0 +1,79 @@
...
+    select
+        'arbitrum' as blockchain, t.*
+    from "query_4420675(blockchain = 'arbitrum', start = '2024-07-29', end = '2100-01-01')" as t
+    union all
+    select
+        'base' as blockchain, t.*
+    from "query_4420675(blockchain = 'base', start = '2024-07-29', end = '2100-01-01')" as t
+)
+
+fselect
```
harisang, line 41: is this a typo?

cowamm/balancer_cow_amms_per_chain_4814818.sql
```diff
@@ -0,0 +1,72 @@
...
+    where (
+        contract_address in ( --the factory contracts
+            0x23fcC2166F991B8946D195de53745E1b804C91B7, --ethereum
+            0xf76c421bAb7df8548604E60deCCcE50477C10462, --ethereum
+            0x23fcC2166F991B8946D195de53745E1b804C91B7, --gnosis
+            0x703Bd8115E6F21a37BB5Df97f78614ca72Ad7624, --gnosis
+            0xe0e2ba143ee5268da87d529949a2521115987302, --arbitrum
+            0x03362f847B4fAbC12e1Ce98b6b59F94401E4588e --base
+        )
+        and topic0 = 0x0d03834d0d86c7f57e877af40e26f176dc31bd637535d4ba153d1ac9de88a7ea
```
harisang, line 18: Nice to have: some reference that confirms this is the topic we are looking for

cowamm/profitability/lp_prices/uni_style_lp_4420675.sql
```diff
@@ -1,93 +1,163 @@
...
 --Parameters
 --  {{blockchain}}: The blockchain to query
 --  {{start}}: The start date of the analysis
 --  {{end}}: The end date of the analysis. date(Timestamp) <= date(timestamp '{{end}}').
 --      For a 1 day period, {{end}} = {{start}}
 
-with date_range as (
+--
+
+with cow_amm_pools as (
```
harisang, line 11: Is it important for performance to only look at Uni pools that have a CoW AMM counterpart?

If so, the comment at the beginning of the query should be updated to explicitly state we are only picking on a subset of queries. Tbh i was surprised to see "cow_amm_pools" as the very first thing in this query and thought it was copy-paste issue

cowamm/profitability/lp_prices/uni_style_lp_4420675.sql
```diff
@@ -1,93 +1,163 @@
...
+            rank() over (partition by contract_address order by day desc) as latest_first
+        from syncs
+        where
+            day < date(timestamp '{{start}}')
+            and latest = 1
+    )
+    where latest_first = 1
+)
+
+select *
```
harisang, line 102: same comment as above

cowamm/profitability/lp_prices/uni_style_lp_4420675.sql
PoloX2021: Yes, we only look at CoW AMM counterparts, otherwise there are 400k pools

cowamm/profitability/lp_prices/uni_style_lp_4420675.sql
```diff
@@ -1,93 +1,163 @@
...
-        and p.token0 = price0.contract_address
-left join prices.day as price1
-    on
-        d.day = price1.timestamp
-        and p.token1 = price1.contract_address
-where
-    (price0.blockchain = '{{blockchain}}' or price0.blockchain is null)
-    and (price1.blockchain = '{{blockchain}}' or price0.blockchain is null)
-    and d.day >= date_trunc('day', p.created_at)
+    from (
```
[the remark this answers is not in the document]
PoloX2021: I wasn't sure about how to do it efficiently. Basically, I want to filter by the final condition (`where latest_first = 1`) AFTER doing all the other operations. If you add the filtering in the where statement, then it happens before which gives a different result. 
If there's another way to do it I'd be glad to learn about it.

cowamm/balancer_cow_amms_per_chain_4814818.sql
PoloX2021: I added a link to an example of bind events & cow amm created events, but I've never really found the documentation about it.

cowamm/balancer_cow_amms_per_chain_4814818.sql
PoloX2021: I don't think it makes it that much more readable no?

cowamm/profitability/lp_prices/uni_style_lp_4420675.sql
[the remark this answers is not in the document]
harisang: aah good point, missed that. Ok, let's leave it as is

harisang (approved):

PR #143, opened 2025-03-05
Competitor analysis

alex-zadorozhnyi:
Added competitor analysis queries to the dune queries repo

### DIFF
--- a/cowprotocol/competitor_analysis/cow_alternatives_analysis_4810844.sql
+++ b/cowprotocol/competitor_analysis/cow_alternatives_analysis_4810844.sql
@@ -1,3 +1,13 @@
+-- Get all users who are using COW Protocol on any chain but {{chain_of_interest}}
+-- Get aggregates (tx_count, usd_volume, user_count) of other dexes/aggregators they use on all chains supported by cow
+-- or limit the results to only competitor usage on {{chain_of_interest}} if {{show_competitors_on_target_chain_only}} is set to 1
+--
+-- Parameters:
+--  {{start_time}} - the trade timestamp for which the analysis should start (inclusive)
+--  {{end_time}} - the trade timestamp for which the analysis should end (inclusive)
+--  {{chain_of_interest}} - which chain is not used by cohort of users we want to investigate
+--  {{show_competitors_on_target_chain_only}} - filters competitor's usage to only {{chain_of_interest}}
+
 with all_cow_users as (
     select
         tx_from as address,
@@ -28,29 +38,26 @@ chains_supported_by_cow as (
         project = 'cow_protocol'
 ),
 
-all_competitor_transactions as (
+trades_dex_and_aggregators as (
     select
-        tx_from as address,
-        blockchain as chain_used_for_competitor,
-        project as competitor_project,
-        sum(amount_usd) as competitor_total_volume_usd,
-        count(*) as competitor_total_transactions
+        tx_from,
+        blockchain,
+        project,
+        amount_usd
     from
         dex.trades
     where
         blockchain in (select blockchain from chains_supported_by_cow)
         and
         block_time between timestamp '{{start_time}}' and timestamp '{{end_time}}'
-    group by 1, 2, 3
 
     union all
 
     select
-        tx_from as address,
-        blockchain as chain_used_for_competitor,
-        project as competitor_project,
-        sum(amount_usd) as competitor_total_volume_usd,
-        count(*) as competitor_total_transactions
+        tx_from,
+        blockchain,
+        project,
+        amount_usd
     from
         dex_aggregator.trades
     where
@@ -59,6 +66,17 @@ all_competitor_transactions as (
         blockchain in (select blockchain from chains_supported_by_cow)
         and
         block_time between timestamp '{{start_time}}' and timestamp '{{end_time}}'
+),
+
+all_competitor_transactions as (
+    select
+        tx_from as address,
+        blockchain as chain_used_for_competitor,
+        project as competitor_project,
+        sum(amount_usd) as competitor_total_volume_usd,
+        count(*) as competitor_total_transactions
+    from
+        trades_dex_and_aggregators
     group by 1, 2, 3
 )
 
--- a/cowprotocol/competitor_analysis/cow_alternatives_analysis_4810844.sql
+++ b/cowprotocol/competitor_analysis/cow_alternatives_analysis_4810844.sql
@@ -1,0 +1,76 @@
+with all_cow_users as (
+    select
+        tx_from as address,
+        array_agg(distinct blockchain) as uses_on_chains
+    from
+        dex_aggregator.trades
+    where
+        project = 'cow_protocol'
+        and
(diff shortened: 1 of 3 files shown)


cowprotocol/competitor_analysis/cow_alternatives_analysis_4810844.sql
```diff
@@ -0,0 +1,77 @@
...
+
+all_competitor_transactions as (
+    select
+        tx_from as address,
+        blockchain as chain_used_for_competitor,
+        project as competitor_project,
+        sum(amount_usd) as competitor_total_volume_usd,
+        count(*) as competitor_total_transactions
+    from
+        dex.trades
```
fleupold, line 39: nit: perform union (with relevant) here to reduce the amount of code duplication in this cte.

cowprotocol/competitor_analysis/cow_alternatives_analysis_4810844.sql
```diff
@@ -0,0 +1,77 @@
+with all_cow_users as (
```
fleupold, line 1: If you look at the other queries in the repo, you will see that we include an explanation as well as a list of arguments for future maintainers.

cowprotocol/competitor_analysis/cow_alternatives_analysis_4810844.sql
```diff
@@ -0,0 +1,77 @@
...
+
+select
+    chain_used_for_competitor,
+    competitor_project,
+    sum(competitor_total_volume_usd) as competitor_total_volume_usd,
+    sum(competitor_total_transactions) as competitor_total_transactions,
+    count(distinct address) as distinct_users_count
+from
+    cow_users_filtered
+left join all_competitor_transactions using (address) -- noqa: disable=L032
```
fleupold, line 73: interesting syntactic sugar.

cowprotocol/competitor_analysis/cow_vs_competition_4807448.sql
```diff
@@ -0,0 +1,101 @@
...
+    from
+        dex_aggregator.trades
+    where
+        project != 'cow_protocol'
+        and
+        blockchain in (select blockchain from chains_supported_by_cow)
+        and
+        block_time between timestamp '{{start_time}}' and timestamp '{{end_time}}'
+    group by 1, 2, 3
+),
```
fleupold, line 71: Instead of duplicating this query, can we express it in terms of the other query (if necessary defining a third shared query, which is used by the other two)?

Cf. mevblocker fee calculation as an example trying to reduce code duplication to a minimum.

fleupold (commented):
Can you document what exactly the purpose of the queries is (it's not clear to me) and think if they can somehow be made dependent on one another (or a third base query) to reduce code duplication?

cowprotocol/competitor_analysis/cow_vs_competition_4807448.sql
alex-zadorozhnyi: I'd prefer to keep it like this until we start seeing that more queries are starting to use the same CTE combo. Refactoring to reduce code duplication would also decrease readability + reduce the ease of forking query in UI to perform ad-hoc analysis.

cowprotocol/competitor_analysis/cow_alternatives_analysis_4810844.sql
alex-zadorozhnyi: guilty, I'm not a big fan of "ON" clause because I don't use table aliases that much and try to keep all the info needed for readability of the code in column aliases

cowprotocol/competitor_analysis/cow_vs_competition_4807448.sql
alex-zadorozhnyi: P.S. I would still do the union before aggregation to reduce that sort of duplication

cowprotocol/competitor_analysis/cow_vs_competition_4807448.sql
```diff
@@ -0,0 +1,101 @@
...
+        blockchain as chain_used_for_cow,
+        count(distinct tx_from) as distinct_users_cow
+    from
+        dex_aggregator.trades
+    where
+        project = 'cow_protocol'
+        and
+        block_time between timestamp '{{start_time}}' and timestamp '{{end_time}}'
+    group by
+        1
```
PoloX2021, line 28: Maybe we could use cow_protocol_target_users here to avoid scannin dex_aggregator.trades again

cowprotocol/competitor_analysis/cow_vs_competition_4807448.sql
```diff
@@ -0,0 +1,118 @@
...
+    select
+        blockchain as chain_used_for_cow,
+        count(distinct tx_from) as distinct_users_cow
+    from
+        dex_aggregator.trades
+    where
+        project = 'cow_protocol'
+        and
+        block_time between timestamp '{{start_time}}' and timestamp '{{end_time}}'
+    group by
```
PoloX2021, line 36: Same here. The time restriction should not be an issue because, all other tables also define a specific timeframe

cowprotocol/competitor_analysis/cow_vs_competition_4807448.sql
```diff
@@ -0,0 +1,118 @@
...
+        chain_used_for_cow,
+        chain_used_for_competitor,
+        competitor_project,
+        count(distinct address) as distinct_users_competitor,
+        sum(total_transactions_on_cow) as transactions_made_on_cow,
+        sum(total_volume_usd_on_cow) as total_volume_usd_on_cow,
+        sum(competitor_total_transactions) as transactions_made_on_competitor,
+        sum(competitor_total_volume_usd) as total_volume_usd_on_competitor
+    from
+        cow_protocol_target_users
```
PoloX2021, line 101: Is it better to do a seperate join for just one additional line, or group it one, with 2 joins

cowprotocol/competitor_analysis/cow_alternatives_analysis_4810844.sql
alex-zadorozhnyi: done

cowprotocol/competitor_analysis/cow_vs_competition_4807448.sql
alex-zadorozhnyi: fixed

alex-zadorozhnyi:
Please review latest version @fleupold @PoloX2021

PoloX2021:
LGTM

cowprotocol/competitor_analysis/cow_vs_competition_4807448.sql
fleupold: I'd really encourage you going forward to think if you can design similar queries (like these two) in ways that share as much as code as possible. You will see that otherwise maintaining a coherent way of doing analysis will become increasingly difficult (e.g. someone may notice a flaw in one of the queries and will want to fix it; now that person needs to somehow know that there are other queries which use the same base logic and would also need to be fixed as otherwise we are starting to see incoherent results).

PR #150, opened 2025-03-12
Add list with auctions excluded from rewards

harisang:
This PR allows the main query that computes rewards to exclude certain auctions from receiving rewards (in cases there is some issue with them)


fleupold (commented):
Do you need this list outside of dune as well? If not you could simply write a query with the auction ids in which the affected order was part of (instead of checking in a 5k lines file under version control)

harisang:
The same list appears in the solver-rewards script as well
https://github.com/cowprotocol/solver-rewards/pull/516/files#diff-f1491bedb3e6d6e90f08b68f90fe3add11635c62391ffc6c8cb74aa38e5a1bbd

harisang:
> Do you need this list outside of dune as well? If not you could simply write a query with the auction ids in which the affected order was part of (instead of checking in a 5k lines file under version control)

Since the order is still active, and potentially other orders might cause this, i thought it would be cleaner to have this list explicitly on github so that we can reference it

cowprotocol/accounting/rewards/main_rewards_dashboard_query_2510345.sql
```diff
@@ -18,36 +18,46 @@ named_results as (
...
+        block_deadline >= (select start_block from block_range)
+        and block_deadline <= (select end_block from block_range)
+),
+
+batch_rewards_filtered as (
+    select
+        br.solver,
+        br.network_fee,
+        br.execution_cost,
+        br.capped_payment * coalesce(ea.multipler, 1) as capped_payment
```
fhenneke, line 42: typo:
```suggestion
        br.capped_payment * coalesce(ea.multiplier, 1) as capped_payment
```

cowprotocol/accounting/rewards/main_rewards_dashboard_query_2510345.sql
```diff
@@ -18,36 +18,46 @@ named_results as (
...
 
 -- AKA Performance Rewards
 primary_rewards as (
     select
-        winning_solver as solver,
+        solver,
         cast(sum(capped_payment) as double) as reward_wei
-    from batch_rewards
+    from batch_rewards_filtered
     group by winning_solver
```
fhenneke, line 52:
```suggestion
    group by solver
```

cowprotocol/accounting/rewards/main_rewards_dashboard_query_2510345.sql
```diff
@@ -18,36 +18,46 @@ named_results as (
...
 fees_and_costs as (
     select
-        winning_solver as solver,
-        cast(sum(fee) as double) as network_fee_wei,
+        solver,
+        cast(sum(network_fee) as double) as network_fee_wei,
         cast(sum(execution_cost) as double) as execution_cost_wei
-    from batch_rewards
+    from batch_rewards_filtered
     group by winning_solver
```
fhenneke, line 61:
```suggestion
    group by solver
```

fhenneke (changes requested):
Testing with https://dune.com/queries/4846013?blockchain_e15077=base&end_time_d83555=2025-03-11+00%3A00%3A00&start_time_d83555=2025-03-04+00%3A00%3A00&quote_cap_native_token_n26d66=0.0002 revealed some typos and missing changes.

harisang:
Thanks! Fixed all of them (sorry for the sloppiness)

fhenneke (approved):
The [modified query](https://dune.com/queries/4846013?blockchain_e15077=base&end_time_d83555=2025-03-11+00%3A00%3A00&start_time_d83555=2025-03-04+00%3A00%3A00&quote_cap_native_token_n26d66=0.0002) does give values with deviations of less than 20 COW (compared to https://github.com/cowprotocol/solver-rewards/pull/516) for the two solvers affected. This should be within normal accuracy of the query.

With regards to code structure, I agree that adding a 5k lines file does not really help much beyond making results reproducible. Since reproducibility is important I approve the PR.

At some point we need a more systematic approach to fixing rewards as the proposed solution would probably not scale well. I do not have a better suggestion, though.

PR #151, opened 2025-03-12
Add tokens symbols to the CoW AMM table [Easy]

PoloX2021:
Adds the token symbols for future display in the dasboards.
Update a token0/1 instead of token1/2 for uniform naming convention


harisang:
>  Update a token0/1 instead of token1/2 for uniform naming convention

I don't see this change here tbh. Seems it was already 0/1

harisang (approved):

PoloX2021:
> > Update a token0/1 instead of token1/2 for uniform naming convention
> 
> I don't see this change here tbh. Seems it was already 0/1

Good catch! I guess I got confused in a query from a previous version before the change

PR #153, opened 2025-03-13
Update again list of excluded auctions

harisang:
Follow-up to PR #152, as another problematic order was identified; see this query: https://dune.com/queries/4847222

**UPDATE:** Actually, PR doesn't work anymore as Dune complains about the length of the query. I edited it by referencing the query  https://dune.com/queries/4847222

**YET ANOTHER UPDATE:** The query 4847222 has now been integrated to the body of the main query


fhenneke:
I do not see a change to the main rewards query, only the removal of the 12k lines file.

harisang:
> I do not see a change to the main rewards query, only the removal of the 12k lines file.

So the results are not hardcoded now but pulled from query [4847222](https://dune.com/queries/4847222). Not sure if that answers your question (?)

fhenneke (approved):
Oh, I see now. The interface on Github mislead me a bit.

Change looks fine for last week.

Going forward, is the expected approach to create a new query for the next exception (like the one modified here) and add that to query 4842868?

harisang:
> Going forward, is the expected approach to create a new query for the next exception (like the one modified here) and add that to query 4842868?

I hope this will be rarely needed, but yeah, this could be an approach.

Do you think i should add that query under version control within this PR as well?

fhenneke:
If you do not think you will extend the query often you can also merge them into one.

If you do not merge, I think both queries should be under version control.

fhenneke:
Looks good.

Issue #124, opened 2025-01-27
Excluded batches for slippage not up to date


fhenneke:
The Dune query https://dune.com/queries/3490353 is out of sync with https://github.com/cowprotocol/dune-queries/blob/main/cowprotocol/accounting/rewards/excluded_batches_query_3490353.sql.

Since this is the case essentially every week, we should think about a process for updating this query.

PR #161, opened 2025-03-31
Update excluded batches from slippage query

harisang:
This query has been updated from time to time ->https://dune.com/queries/3490353
but the changes are not reflected in the code in this repo (see also relevant issue #124 )

This PR brings the code up-to-date, and we commit to only edit the query via PRs in this repo moving forward


harisang:
To review this PR, one only needs to check that this file
https://github.com/cowprotocol/dune-queries/blob/2af41cc15a63be9cba64d2c62773e7b1e5983f26/cowprotocol/accounting/rewards/excluded_batches_query_3490353.sql

is identical to query here->https://dune.com/queries/3490353

cowprotocol/accounting/rewards/excluded_batches_query_3490353.sql
```diff
@@ -168,3 +169,62 @@ where
...
+    or tx_hash = 0x50D5F477B7F62BF2F4CB258E9D5AC375033DF56F4C0CC07DE4E9C585D90AA390
+
+    -- for week of Feb 25 - March 4, 2025 on mainnet
+    or tx_hash = 0x34AD176862C6D51BDD8232B67BF4C689CF0707CEC18E17E9731017A1F3823F6A
+
+    -- for week of March 4 - March 11, 2025 on mainnet
+    or tx_hash = 0x1F14B47CCA77FBA18F6AD87372C9F942617FAF9C787622BBD8F56D187EF069FA
+
+-- Base
+union all
```
fhenneke, line 214: The line on Dune is
```suggestion
union
```
which is a `union distinct`.
It should not make a difference but also does not look like an error from copy-pasting.

bram-vdberg (approved):

PR #163, opened 2025-04-01
User segmentation

alex-zadorozhnyi:
Adding user segmentation dashboard under VCS. Next iteration would require adding layered transactions, but will put it into next PR

### DIFF
--- a/cowprotocol/user_segmentation/user_segmentation_4872780.sql
+++ b/cowprotocol/user_segmentation/user_segmentation_4872780.sql
@@ -17,7 +17,7 @@
 
 with creation_month_per_user as (
     -- TODO: use both dex and dex-agg here
-    select distinct
+    select
         taker as user_address,
         min(date_trunc('Month', block_date)) as creation_month
     from
@@ -26,6 +26,7 @@ with creation_month_per_user as (
         blockchain = '{{chain_name}}'
     group by 1
 ),
+
 all_active_users as (
     -- TODO: use layered txs here to prevent miss-information
     select distinct
@@ -39,6 +40,7 @@ all_active_users as (
         and
         block_date >= timestamp '{{start_time}}' - interval '1' month
 ),
+
 users_backbone as (
     select
         user_address,
@@ -46,21 +48,21 @@ users_backbone as (
         null as project
     from
         all_active_users
-    cross join unnest(sequence(date_trunc('Month', timestamp '{{start_time}}') - interval '1' month, current_date, interval '1' month)) as a(dt)
+    cross join unnest(sequence(date_trunc('Month', timestamp '{{start_time}}') - interval '1' month, current_date, interval '1' month))
 ),
+
 all_users as (
-    select
-        *
+    select *
     from
         users_backbone
 
     union all
 
-    select
-        *
+    select *
     from
         all_active_users
 ),
+
 user_w_creation_date as (
     select
         user_address,
@@ -69,10 +71,11 @@ user_w_creation_date as (
         project
     from
         all_users
-    left join creation_month_per_user using (user_address)
+    left join creation_month_per_user on all_users.user_address = creation_month_per_user.user_address
     where
         creation_month <= block_month
 ),
+
 users_labeled as (
     select
         user_address,
@@ -88,6 +91,7 @@ users_labeled as (
         user_w_creation_date
     group by 1, 2
 ),
+
 users_w_last_month_status as (
     select
         user_address,
--- a/cowprotocol/user_segmentation/user_segmentation_4872780.sql
+++ b/cowprotocol/user_segmentation/user_segmentation_4872780.sql
@@ -1,0 +1,118 @@
+-- Query that calculates for each [user, cal. month] pair a user status that is one of ['Loyal', 'Opportunistic', 'Lost', 'Gone', 'Didnt exist']
+-- Then it aggregates it by month and calculates transitions from one status to another month over month.
+-- It is parametrised by chain and project.
+-- Status descriptions:
+-- Loyal - only traded {{project}} during calendar month
+-- Opportunistic - traded both {{project}} during calendar month and other projects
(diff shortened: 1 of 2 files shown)


harisang:
I checked the code here vs the actual query on Dune and there are some small differences

```
2c2
<     select
---
>     select distinct
32c32
<     cross join unnest(sequence(date_trunc('Month', timestamp '{{start_time}}') - interval '1' month, current_date, interval '1' month))
---
>     cross join unnest(sequence(date_trunc('Month', timestamp '{{start_time}}') - interval '1' month, current_date, interval '1' month)) as a(dt)
57c57
<     left join creation_month_per_user on all_users.user_address = creation_month_per_user.user_address
---
>     left join creation_month_per_user using (user_address)
```

Was this intended?

alex-zadorozhnyi:
@harisang  Good catch! These linter's changes which I've missed. Pushing fix now!

cowprotocol/user_segmentation/user_segmentation_4872780.sql
```diff
@@ -0,0 +1,122 @@
...
+        users_labeled
+)
+
+
+select
+    block_month,
+    previous_status || '->' || current_status as status_shift,
+    count(distinct user_address) as users_count,
+    '{{current_status}}' as debug_1,
+    '{{previous_status}}' as debug_2
```
fleupold, line 112: do you wnat to keep these?

cowprotocol/user_segmentation/user_segmentation_4872780.sql
```diff
@@ -0,0 +1,122 @@
...
+        case when project in ('1inch', '1inch Limit Order Protocol') then '1inch' else project end as project
+    from
+        dex_aggregator.trades
+    where
+        blockchain = '{{chain_name}}'
+        and
+        block_date >= timestamp '{{start_time}}' - interval '1' month
+),
+
+users_backbone as (
```
fleupold, line 44: maybe a small inline comment why this is needed (not super clear from the name and the sequence is not trivial to understand)

cowprotocol/user_segmentation/user_segmentation_4872780.sql
```diff
@@ -0,0 +1,122 @@
...
+    select
+        all_users.user_address,
+        block_month,
+        creation_month,
+        project
+    from
+        all_users
+    left join creation_month_per_user on all_users.user_address = creation_month_per_user.user_address
+    where
+        creation_month <= block_month
```
fleupold, line 76: Is this where clause needed? Shouldn't creation month always be smaller or equal than block month unless we have a logic error somewhere?

cowprotocol/user_segmentation/user_segmentation_4872780.sql
```diff
@@ -0,0 +1,122 @@
...
+        block_month,
+        any_value(creation_month) as creation_month,
+        case
+            when count(distinct project) = 1 and max(project) = '{{project}}' then 'Loyal'
+            when contains(array_agg(project) filter (where project is not null), '{{project}}') then 'Opportunistic'
+            when count(distinct project) >= 1 and not contains(array_agg(project) filter (where project is not null), '{{project}}') then 'Lost'
+            else 'Gone'
+        end as user_type
+    from
+        user_w_creation_date
```
fleupold, line 91: nit: could join `creation_month_per_user` here to save one CTE and make the query a bit shorter

cowprotocol/user_segmentation/user_segmentation_4872780.sql
```diff
@@ -0,0 +1,122 @@
...
+users_labeled as (
+    select
+        user_address,
+        block_month,
+        any_value(creation_month) as creation_month,
+        case
+            when count(distinct project) = 1 and max(project) = '{{project}}' then 'Loyal'
+            when contains(array_agg(project) filter (where project is not null), '{{project}}') then 'Opportunistic'
+            when count(distinct project) >= 1 and not contains(array_agg(project) filter (where project is not null), '{{project}}') then 'Lost'
+            else 'Gone'
```
fleupold, line 88: I personally find this classification hard to follow. Could we maybe simplify it by

- when count = 0 => Gone
- when count => 1 and not(contain(project)) => Lost
- when count = 1 => loyal
- else => opportunistic

Assuming the branches are not pass through.

cowprotocol/user_segmentation/user_segmentation_4872780.sql
```diff
@@ -0,0 +1,122 @@
...
+    group by 1, 2
+),
+
+users_w_last_month_status as (
+    select
+        user_address,
+        block_month,
+        creation_month,
+        user_type as current_status,
+        lag(user_type, 1, if(creation_month >= block_month, 'Didnt Exist', 'Gone')) over (partition by user_address order by block_month asc) as previous_status
```
fleupold, line 101: Can the Gone case be hit here (now that we have a full range of months for each user)? If not, let's just remove the if

cowprotocol/user_segmentation/user_segmentation_4872780.sql
```diff
@@ -0,0 +1,122 @@
...
+--  {{project}} - project to do the analysis for
+--  {{current_status}} - filter to limit analysis to users that had this status as current for given calendar month
+--  {{previous_status}} - filter to limit analysis to users that had this status as previous for given calendar month
+
+
+with creation_month_per_user as (
+    -- TODO: use both dex and dex-agg here
+    select
+        taker as user_address,
+        min(date_trunc('Month', block_date)) as creation_month
```
fleupold, line 22: nit: might want to make the aggregate by configurable

cowprotocol/user_segmentation/user_segmentation_4872780.sql
alex-zadorozhnyi: yeah, dune UI seems laggy sometimes when multichoice list is used

PR #165, opened 2025-04-02
[MEV Blocker] Add new builder btcs

alex-zadorozhnyi:
Adding new builder, as requested by Alex V.


fleupold (approved):
Why are there so many builder addresses?

PR #167, opened 2025-04-08
[MEV Blocker] rollback fee reduction

alex-zadorozhnyi:
Removed temporary fee reduction. We want fee to become 50%, before it was artificially reduced to 40%.


harisang (approved):

fleupold (commented):
Looks good!

PR #168, opened 2025-04-15
Exclude tx with scam token

harisang:
Garbage token that was denylisted, where owners could transfer out from the settlement contract (see [here](https://etherscan.io/tx/0x6259db0f1bac7ecc9b69cb35ea3b6432c799ce6031c85abefa62518ba22e530f))


bram-vdberg (approved):
LGTM

PR #170, opened 2025-05-04
Exclude barn flashloans tx from slippage

harisang:
Proper indexing of flashloans queries is not yet completed on Dune, so the [test tx](https://etherscan.io/tx/0x683fff5607fac2fd249050794fedd47875e3f94762ee885c7514b080787989c5) on mainnet barn that was executed last week currently creates issues with the payouts script. Because of that, we decided to exclude it from slippage accounting. It is easy to see that slippage in the tx is essentially zero.

Change has been already applied.


bram-vdberg:
LGTM

bram-vdberg (approved):

PR #171, opened 2025-05-06
Update partner fees

harisang:
This PR updates the partner fees query, based on the current state of the config file of the payouts script:
https://github.com/cowprotocol/solver-rewards/blob/a2b7a7b44c6b6880c44372224f5f0ec0d5b7cf44/src/config.py#L116

It also removes some duplication of addresses


bram-vdberg (approved):
LGTM

PR #173, opened 2025-05-14
Prevent Duplicate Counting in Solver Metrics by Pre-Aggregating Trades

sissnad:
### Overview

This PR fixes a data integrity issue in the `V2: Ranked Active Solvers` Dune query, where solver metrics were **inflated due to row duplication** caused by directly joining `batches` with `trades` using `tx_hash`.

### What was wrong?

- The original query included:
  ```sql
  INNER JOIN trades t ON b.tx_hash = t.tx_hash
  ```
  Since many batches have multiple trades, this caused row multiplication, leading to overcounting.
  
 ### What’s changed?
- Added a trade_agg CTE to aggregate trade metrics per tx_hash.

- Joined batches to this pre-aggregated view to avoid row duplication.

- Ensured only settled batches (those with at least one trade) are included by joining against trade_agg.

- Preserved solver-level filters (environment NOT IN ('test', 'service'), active = TRUE).

### DIFF
--- a/cowprotocol/solver_dashboard/ranked_active_solvers_1756858.sql
+++ b/cowprotocol/solver_dashboard/ranked_active_solvers_1756858.sql
@@ -3,32 +3,59 @@
 -- Parameters:
 --   {{last_n_days}}: int the number of days to look back
 
-with solver_info as (
-    select
-        name as solver_name,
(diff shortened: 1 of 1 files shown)


matsai5 (approved):
Looks correct. Thank you.

PR #180, opened 2025-06-03
Add slippage transactions for the week of May 27

bram-vdberg:


bram-vdberg:
Dune lists the price for $DUCK at `0.002537` USD for this timestamp, but the price was actually around `0.0002281` USD per token.

PR #183, opened 2025-06-09
Update monthly_dao_revenue_3700123.sql

### LINKED ISSUE #181: Fix gnosis section of monthly revenue query
In [this query](https://github.com/cowprotocol/dune-queries/blob/3208541da3de6f43205df454cca84c55b4fcfdeb/cowprotocol/revenue/monthly_dao_revenue_3700123.sql), in the CTE for gnosis, we have blockchain='arbitrum'. 
Fixing this would involve not only adjusting for blockchain='gnosis' but also transforming the native token into ETH, as all other networks' native tokens are ETH.

fleupold:
fixes #181


matsai5:
Since this query is calculating revenue in ETH, using the same logic (native token price) is not enough for gnosis chain. We must add a conversion from xdai to eth too.

matsai5 (approved):

PR #185, opened 2025-06-11
Use network fees from raw order data whenever available

harisang:
We are about to start uploading explicitly `network_fee` amounts in the raw order data table, while deprecating the `surplus_fee` column at the same time. This PR ensures that once the `network_fee` column appears in the raw order data table, all queries that have used `surplus_fee` up till now will continue to work.


harisang:
@fhenneke sorry had forgotten to attach the test query

https://dune.com/queries/5289093?end_time_d83555=2025-06-19+00%3A00%3A00&start_time_d83555=2025-06-10+00%3A00%3A00

PR #186, opened 2025-06-11
Add auction data query

harisang:
This PR introduces a new query that provides some info on a per auction basis about each winning solver, the total network fees, execution costs and protocol fees collected from all its winning solutions that landed onchain, its relevant offchain score (named `competition_score`), its `observed_score`, its `reference_score` and the corresponding payment (uncapped and capped).

Query can be found here: https://dune.com/queries/5270914

**Note:** the query uses the assumption that the raw_batch_data table that we sync on Dune contains a mix of solution and solver information. Specifically, it is assumed that the `reference_score`, the `uncapped_payment_native_token` and the `capped_payment` entries all refer to the (auction_id, solver) pair and thus are identical for all different solutions the solver won.


cowprotocol/accounting/rewards/auction_data_5270914.sql
```diff
@@ -0,0 +1,64 @@
...
+--    auction_id: integer
+--    block_deadline: integer
+--    solver: varbinary
+--    total_network_fee: decimal(38, 0)
+--    total_execution_cost: decimal(38, 0)
+--    total_protocol_fee: decimal(38, 0)
+--    competition_score: decimal(38, 0)
+--    observed_score: decimal(38, 0)
+--    uncapped_payment_native_token: decimal(38, 0)
+--    capped_payment_native_token: decimal(38, 0)
```
fhenneke, line 17: Good idea to annotate output types.

How did you check the types here?

cowprotocol/accounting/rewards/auction_data_5270914.sql
```diff
@@ -0,0 +1,64 @@
...
+)
+
+select --noqa: ST06
+    rbd.environment,
+    rbd.auction_id,
+    rbd.block_deadline,
+    rbd.solver,
+    sum(rbd.network_fee) as total_network_fee,
+    sum(coalesce(rbd.execution_cost, txs.execution_cost, 0)) as total_execution_cost,
+    sum(rbd.protocol_fee) as total_protocol_fee,
```
fhenneke, line 43: Does this handle cases of partial reverts correctly?

I would have expected a more uniform treatment of execution costs and fees. E.g. either make all zero in case of (partial) reverts, or make them all null in case of full reverts. At the moment, in case of full reverts, costs are zero and fees are null.

cowprotocol/accounting/rewards/auction_data_5270914.sql
```diff
@@ -0,0 +1,64 @@
...
+    on rbd.environment = txs.environment and rbd.auction_id = txs.auction_id and rbd.block_deadline = txs.block_deadline and rbd.solver = txs.solver
+where rbd.block_deadline >= (select start_block from block_range) and rbd.block_deadline <= (select end_block from block_range)
+group by
+    rbd.environment,
+    rbd.auction_id,
+    rbd.block_deadline,
+    rbd.solver,
+    rbd.reference_score,
+    rbd.uncapped_payment_native_token,
+    rbd.capped_payment
```
fhenneke, line 64: I would add a comment here, similar to what you describe in the test of the PR.

cowprotocol/accounting/rewards/auction_data_5270914.sql
harisang: Ahh good point. My intention was to have it explicit, similar to how it's done in the raw order/batch data queries, so that we know how to handle those. But i admit i haven't checked the output types, and i am not sure i know how to do it. The only way to go for me would be to explicitly add type casing in the final `select` statement.

harisang:
Looks good! @fhenneke you need to approve as was authored initially by me

fhenneke (approved):

PR #190, opened 2025-06-22
Remove deprecated bonding pools

harisang:
This PR removes bonding pools that are not valid anymore from the relevant dune queries, so as to reduce the possibility of side effects in the accounting pipeline


harisang:
Closing this without merging as removing those entries will make it impossible to recover old vouchings.

PR #196, opened 2025-06-26
New partner fee splits query

matsai5:


fleupold (commented):
structure looks good to me leaving it to @cowprotocol/solver to double check that the list is complete and correct

harisang (approved):

PR #198, opened 2025-07-07
Add avalanche and polygon support to multiple accounting-related queries

harisang:
This PR adds Avalanche and Polygon support to several queries we are using for accounting-related tasks


fhenneke (approved):
The changes to the main dashboard look good.

I cannot comment on the other changes. (Are we using the simplified approach here, ignoring app codes? Can we start using that in the rest of the accounting as well?)

PR #200, opened 2025-07-08
added blocks data to use partitioning in tx

data-cowwboy:
While filtering on transactions, we should try to involve block_date whenever we can. This is because the transactions (logs,traces) tables are partitioned on block_date and it helps avoid looking into the full table if we add it in the filter

### DIFF
--- a/cowprotocol/accounting/rewards/auction_data_5270914.sql
+++ b/cowprotocol/accounting/rewards/auction_data_5270914.sql
@@ -29,7 +29,11 @@ txs_block_range as (
 ),
 
 block_data as (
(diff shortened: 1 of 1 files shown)


fhenneke:
> @fhenneke can you please test again. The new version is giving me better performance each time. Did you use the same query engine size for both ?

I am using the cowprotocol account and running using the medium engine. I repeated the experiment today and the old query took 54 seconds, the old one 2 minutes. Are you running the queries via my link (on the same time period)? What numbers do you get?

fhenneke (approved):
Approving as the change looks conceptually correct and does not seem to mess with results.

The questions about benchmarking were more out of curiosity.

harisang (approved):

Issue #191, opened 2025-06-23
Slippage computation wrong for transfers from the settlement contract to itself


fhenneke:
To handle fee withdrawals correctly, transfers from the settlement contract to itself are counted as increasing buffers. This is wrong if solutions include additional transfers of that form. One example seems to be due to flash loans, which automatically generate transfers of that form.

The slippage query needs to be adapted to handle such transfers correctly.

PR #201, opened 2025-07-09
[Bug] Correct buffer accounting

fhenneke:
## Motivation

This PR fixes the bug from #191.

The issue was with how we correct buffer imbalances in case of fee withdrawals. In these settlements, buffers decrease as intended. But we do not want to misclassify that as negative slippage. We therefore included a rule that all transfers from the settlement contract to itself are treated as if balances had increased. That took care of the case of fee withdrawals, which have such transfers. It did not correctly handle other cases of such transfers. One such case is with the current implementation of flashloans.

## Changes

This PR changes two queries.
* Query 4021644 on imbalances does not correct for fee withdrawals anymore. Instead it correctly computes changes of buffers now.
* Query 4059683 uses the previous query but now introduces special handling of trades from the settlement contract to reward safes.

In combination, fee withdrawals are still treated the correct way, flash loans are now treated correctly, and other transfers from the settlement contract to itself are ignored.

## Testing

I compared results of the new and old query for a one-day period including a flash loan transaction. Test queries are here:
* [Old imbalances](https://dune.com/queries/4021644?end_time_d83555=2025-06-11+00%3A00%3A00&start_time_d83555=2025-06-10+00%3A00%3A00)
* [New imbalances](https://dune.com/queries/5436278?end_time_d83555=2025-06-11+00%3A00%3A00&start_time_d83555=2025-06-10+00%3A00%3A00)
* [Old slippage per transaction](https://dune.com/queries/4059683?raw_slippage_table_name_e15077=raw_slippage_per_transaction&end_time_d83555=2025-06-11+00%3A00%3A00&start_time_d83555=2025-06-10+00%3A00%3A00)
* [New slippage per transaction](https://dune.com/queries/5436824?end_time_d83555=2025-06-11+00%3A00%3A00&start_time_d83555=2025-06-10+00%3A00%3A00&raw_slippage_table_name_e15077=raw_slippage_per_transaction)

Restricting results to the transactions a [flash loan settlement](https://etherscan.io/tx/0x0183756d30137630a9c1f7c02c9ec904751147e0fedbd7e529f31d05aac04baf) and a [fee withdrawal](0x799e6d637b9df6b4a659e9687d8509aebebb58ca34392fea3f6ad2e37e5e44a7).

Looking at the final slippage per transaction, we see that the slippage for the flashloan transaction changes from around $1k to around $0.01. This corresponds to actual slippage of the transaction.

The result for the fee withdrawal unchanged (up to rounding errors).

With the query for imbalances one can observe a change for both transactions. The flash loan transaction now shows almost no slippage, while the fee withdrawal shows the actual decrease in buffers.

## Limitations

There are other ways the settlement contract can appear as trader. For example an external trader could create an order with the settlement contract being the receiver of funds. This case is (probably) not handled correctly with this PR (nor with the current approach). One could at least have expected that buffers increase when settling such trades. Instead, with this PR, it will count towards positive slippage of a solver.

Since this case is rare, I decided against implementing special handling of such trades. One practical issue was that I cannot just look at trades where the settlement contract is the receiver since that would also lead to corrections for flash loans. So there would have to be another correction for flashloans. But the flash loans implementation is already about to be changed.

### DIFF
--- a/cowprotocol/accounting/slippage/raw_slippage_4059683.sql
+++ b/cowprotocol/accounting/slippage/raw_slippage_4059683.sql
@@ -14,8 +14,11 @@
 --   0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee for native token
 -- - amount: signed value of slippage in atoms of the token; fees are represented as negative
 --   integers since they will be removed from imbalances
--- - slippage_type: 'raw_imbalance' for imbalance observable on chain, 'protocol_fee' for the total
---   protocol fee (including partner fee), 'network_fee' for network fees
+-- - slippage_type:
+--     'raw_imbalance' for imbalance observable on chain,
+--     'protocol_fee' for the total protocol fee (including partner fee),
+--     'network_fee' for network fees,
+--     'settlement_contract_sell' for corrections due to fee withdrawals
 -- - price_unit: USD price of one unit (i.e. pow(10, decimals) atoms) of a token
 -- - price_atom: USD price of one atom (i.e. 1. / pow(10, decimals) units) of a token
 -- - slippage_usd: USD value of slippage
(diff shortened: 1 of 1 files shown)


cowprotocol/accounting/slippage/raw_slippage_4059683.sql
```diff
@@ -49,10 +52,33 @@ fees as (
...
+        tx_hash,
+        sell_token_address as token_address,
+        atoms_sold as amount,
+        'settlement_contract_sell' as slippage_type,
+        date_trunc('hour', block_time) as hour --noqa: RF04
+    from cow_protocol_{{blockchain}}.trades
+    where
+        block_time >= timestamp '{{start_time}}' and block_time < timestamp '{{end_time}}'
+        and trader = 0x9008D19f58AAbD9eD0D60971565AA8510560ab41
+        and receiver = case
```
harisang, line 70: I do find this a bit "risky" as it means we need to edit this query whenever we deploy on a new chain or change the receiver of the fee withdrawals.

Asked the backend whether we accept such orders; if not, then the case would have in mind here would lead to blacklisting so maybe we don't need to take it into account

harisang: Given the discussion with the backend, this condition is actually needed.

Given that rewards safes show up in this query as well -> https://dune.com/queries/2601653
should we introduce a new query that just lists those, so that we have one place to maintain?

fhenneke: We should probably do that. But in a separate PR.

PR #202, opened 2025-07-10
Update slippage query for current pipeline

fhenneke:
This PR adds a change to the raw splippage query to produce an output compatible with our data pipeline.

That changed version is already used in production via [this query](https://dune.com/queries/5143874). This PR allows for switching our pipeline to use the query under version control.

If significant changes are required to the query, they should be done in a follow up PR.

## Context

The reason for the change is that the incremental strategy used in our accounting requires a unique key. Without the grouping, the combination of columns `(tx_hash, token_address, slippage_type)` is not unique. With this change it is.

The problem is that in principle there can be identical rows in the breakdown. The grouping forces it to become unique, without much interference with the rest of the code.

One piece of information which is lost with this approach is multiple protocol fees on the same token.

## Testing

The proposed change is already used in production.


harisang (approved):

harisang:
@fhenneke Should we merge this?

PR #219, opened 2025-07-24
Lookup tables

matsai5:
This PR introduces a new folder for lookup tables, i.e, hardcoded data that is often referenced in our queries but is not currently kept in a single place. This will make our code simpler and easier to maintain.


cowprotocol/lookup_tables/daily_stable_tokens_5469456.sql
```diff
@@ -0,0 +1,41 @@
...
+with
+daily_prices as (
+    select
+        date_trunc('day', timestamp) as date,
+        contract_address as token_address,
+        blockchain,
+        min(price) as daily_min_price -- to avoid temporary pricing errors
+    from prices.hour
+    where
+        timestamp >= timestamp '2022-01-01'
```
harisang, line 11: What is the purpose of this query? Asking as i am curious to see if it makes sense to start looking that back in time? And not, for example, look at the past year

cowprotocol/lookup_tables/daily_stable_tokens_5469456.sql
```diff
@@ -0,0 +1,41 @@
+-- this query returns a daily list of token addresses which fit into a few price stability criteria over the previous 30d
```
harisang, line 1: could you maybe describe these stability criteria (if this is easy) up here so that there is no need to look at the code?

cowprotocol/lookup_tables/valid_reward_targets_5492689.sql
```diff
@@ -0,0 +1,43 @@
...
+
+union distinct
+select distinct cowRewardTarget--, 'avalanche_c' as blockchain
+from cow_protocol_avalanche_c.VouchRegister_evt_Vouch
+inner join query_4056263 -- valid bonding pools
+    on pool_address = bondingPool
+    and sender = creator
+
+/*
+union distinct
```
harisang, line 37: due to issues with namespaces for Polygon (which we should ping the Dune team about), as a workaround we have this query that collects all vouching events for all networks

https://dune.com/queries/5513654

harisang (changes requested):

cowprotocol/lookup_tables/daily_stable_tokens_5469456.sql
matsai5: The purpose is to be able to flag which trades are stable-stable, i.e., impossible to generate significant revenue on. Since our protocol fee started in 2024 I'll change it to that date.

harisang:
@matsai5 I saw you started using this query to track vouching events.
https://dune.com/queries/5533098

Definitely fine with this, but we should decide if we want to keep it materialized, and if so, decide on a schedule. Also it is not currently under version control here (i think) so could you add it in a follow-up PR.

Other than that, things look good

harisang:
Will approve now but could you add a description as well in the PR before merging?

harisang (approved):

Issue #207, opened 2025-07-16
Add all queries for new accounting pipeline


fhenneke:
We currently use several queries in our accounting pipeline which are not under version control.

- [x] 5143848, 5143758 for vouches (#212)
- [ ] 5143862 for solvers
- [x] 5143874 for slippage (#202)
- [ ] 5143916 for service fee
- [ ] 5143918 for block times
- [ ] 4545357 for conversion rates (see also #222)

Most of them just use a sligthly different format from the original query.

They should all be added to this repo, or the calling code should be refactored to use queries from this repo.

PR #222, opened 2025-07-28
Add conversion prices query

harisang:
This PR introduces a materialized conversion prices query that precomputes the conversion prices (COW/native token) used in the weekly accounting. This is meant to speed up the main solver rewards query.

New query: https://dune.com/queries/5533118

Schedule: materialize on Tuesdays at 4am.

Test query: https://dune.com/queries/5544153

Note: i noticed test query might return non-empty results, but only the very last decimal might differ, so this must be due to rounding differences

image


fhenneke:
Should this query replace https://dune.com/queries/4545357 in the dagster pipeline? The format is a bit different, but these queries are used for the same purpose.

fhenneke:
I will link this PR in #207 .

PR #238, opened 2025-08-14
Dex agg top layer metrics

matsai5:


cowprotocol/competitor_analysis/dex_aggregator_layered_metrics_5616287.sql
```diff
@@ -0,0 +1,100 @@
...
+            *
+            , min(proj_trace_address) over (partition by tx_hash) as tx_min_trace_address
+        from dex_agg_trades_prep_2
+    )
+    where proj_trace_address = tx_min_trace_address
+)
+, dex_agg_trades as (
+    select 
+        date_trunc('{{date_granularity}}', block_time) as date
+        --, blockchain
```
fleupold, line 78: nit: remove for cleanup

cowprotocol/competitor_analysis/dex_aggregator_layered_metrics_5616287.sql
```diff
@@ -0,0 +1,100 @@
...
+-- for each tx, only keep project with shallowest trace_address
+, dex_agg_trades_prep_2 as (
+    select 
+        block_time
+        , blockchain
+        , tx_hash
+        , project
+        , sum(amount_usd) as volume
+        , count(distinct trader) as traders
+        , count(1) as trades
```
fleupold, line 60: I don't think this correctly accounts for our batches (taker will be the settlement contract and we can have multiple trades in one settlement).

cowprotocol/competitor_analysis/dex_aggregator_layered_metrics_5616287.sql
```diff
@@ -0,0 +1,100 @@
...
+        t.tx_hash,
+        t.evt_index,
+        t.trace_address,
+        lower(coalesce(f.project, t.project)) as project, 
+        t.amount_usd,
+        t.token_sold_symbol,
+        t.token_bought_symbol,
+        t.token_sold_amount,
+        t.token_bought_amount,
+        t.taker as trader
```
fleupold, line 38: I believe you can:

```suggestion
        t.taker as trader,
        row_number() over (
              partition by t.tx_hash
              order by trace_address, evt_index
        ) as rn
```

This way you don't need the other prep tables and can just filter on `where rn = 1` in the final CTE. Not sure if this has performance implications though.

cowprotocol/competitor_analysis/dex_aggregator_layered_metrics_5616287.sql
```diff
@@ -0,0 +1,100 @@
+-- Calculates metrics based on the project of the top layer of a transaction, regardless of whether its routing touches other dex aggregators
+with
+static as (
+    select 
+        split('{{blockchain}}', ',') as blockchains
+        , '{{date_granularity}}' as date_granularity
+        , (select start_date['{{period}}'] from "query_5633784") as start_date
```
fleupold, line 7: Nice! Should we also put this under version control?

cowprotocol/competitor_analysis/dex_aggregator_layered_metrics_5616287.sql
```diff
@@ -0,0 +1,100 @@
...
+        , '1inch fusion' as project        
+    from oneinch.swaps, static
+    where
+        flags['fusion'] 
+        and block_time >= static.start_date
+        and if(array_position(static.blockchains, '-=All=-') > 0, true, array_position(static.blockchains, blockchain) > 0) 
+)
+, uniswap_x_txs as ( -- some uniswap X txs are present in "dex_aggregator.trades", we'll first filter them out to avoid double counting
+    select *
+    from "query_5602624(period='{{period}}')", static
```
fleupold, line 22: This one should definitely be versioned controlled given its complexity.

fleupold (approved):
Main correctness point on how we break ties on traceAddress as well as not relying on the trader and trade numbers for CoW (which don't account for batching).

Otherwise looks great!

cowprotocol/competitor_analysis/dex_aggregator_layered_metrics_5616287.sql
```diff
@@ -0,0 +1,103 @@
...
+    select 
+        date_trunc(date_granularity, block_time) as date
+        , project
+        , sum(amount_usd) as volume
+        , count(*) as trades
+        , count(distinct trader) as traders
+    from dex_agg_trades_prep_3
+    group by 1,2
+)
+-- Add uniswap X on top since it's always 1st layer
```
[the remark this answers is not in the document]
matsai5: Uniswap x is not present as a project in in dex_aggregator.trades, and given it's also intent-based, we can assume it will be 1st layer always. The reason to exclude uniswap x tx hashes from dex_aggregator.trades is because some txs from uniswap x do appear in that table but under other projects, due to routing. So we remove them to avoid double counting.

PR #244, opened 2025-09-02
Add prices to mainnet price feed for August 29 - Sep2, 2025 accounting

harisang:
It was flagged that these 2 txs don't show any slippage on Dune, due to missing prices.

https://etherscan.io/tx/0x9cab916a0766d8a2b3f235f29766923d51b50a58f44ef5c4d6a2cbeba669e600

https://etherscan.io/tx/0x31a5fa7e29c644563ef583d641d851b1689780882c107b4799b5a4a1fb0f3a35

One can verify this by looking at this query:
https://dune.com/queries/4059683?end_time_d83555=2025-09-02+00%3A00%3A00&raw_slippage_table_name_e15077=raw_slippage_per_transaction&start_time_d83555=2025-08-26+00%3A00%3A00

and filtering by tx hash

The reason for this is due to missing prices of the buy tokens in the relevant hour time window when each tx took place, which can be verified here:
https://dune.com/queries/4064601?blockchain_t6c1ea=ethereum&end_time_d83555=2025-09-02+00:00:00&start_time_d83555=2025-08-26+00:00:00

To fix this, we looked at the relevant auctions
- https://api.cow.fi/mainnet/api/v2/solver_competition/by_tx_hash/0x9cab916a0766d8a2b3f235f29766923d51b50a58f44ef5c4d6a2cbeba669e600
- https://api.cow.fi/mainnet/api/v2/solver_competition/by_tx_hash/0x31a5fa7e29c644563ef583d641d851b1689780882c107b4799b5a4a1fb0f3a35

and computed the exchange rate between the buy token and WETH, and then used that exchange rate combined with the WETH price at the hour of each trade to get a USD value for each token.

Concretely,
- 0x4d5f47fa6a74757f35c14fd3a6ef8e3c9bc514e8: exchange rate is 1 to WETH
- 0x4d5f47fa6a74757f35c14fd3a6ef8e3c9bc514e8: exchange rate is `1208909705490099200 /10**18 = 1.2089097054900992`

The relevant prices follow from the above exchange rates and the prices shown in the screenshots below

image
image

### DIFF
--- a/cowprotocol/accounting/slippage/slippage_prices_4064601.sql
+++ b/cowprotocol/accounting/slippage/slippage_prices_4064601.sql
@@ -37,9 +37,9 @@ precise_prices as (
         prices.usd
     inner join token_times
         on
-            date_trunc('hour', minute) = hour
-            and contract_address = token_address
-            and blockchain = '{{blockchain}}'
+        date_trunc('hour', minute) = hour
+        and contract_address = token_address
+        and blockchain = '{{blockchain}}'
     group by 1, 2, 3
 ),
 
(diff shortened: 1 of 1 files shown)


fhenneke (approved):
I checked the WETH price with coingecko. I did not check the wstETH price, but the logic sounds reasonable.

I assume this is an issue with all Aave tokens, right?
We might need to change our approach to prices for new tokens, as the old prices table does not seem to be updated. Unfortunately, the new one is also not a replacement.

PR #250, opened 2025-09-09
Adapt price feed in slippage accounting for Aave tokens

harisang:
As the `prices.usd` Dune table is not working great with Aave tokens, this query uses a mapping of these tokens to the underlying asset in order to use the address of the underlying asset for recovering the price of the token itself.

The mapping is done using this cte

```
aave_tokens_mapping as (
    select distinct
        chain as blockchain,
        atoken as atoken_address,
        asset as underlying_address
    from aave_v3_multichain.poolconfigurator_evt_reserveinitialized
    union distinct
    select distinct
        'ethereum' as blockchain,
        atoken as atoken_address,
        asset as underlying_address
    from aave_v3_lido_ethereum.poolconfigurator_evt_reserveinitialized
    union distinct
    select distinct
        'ethereum' as blockchain,
        atoken as atoken_address,
        asset as underlying_address
    from aave_v3_ethereum.etherfipoolconfiguratorinstance_evt_reserveinitialized
),
```

Modified query can be tested here: https://dune.com/queries/5749995

### DIFF
--- a/cowprotocol/accounting/slippage/slippage_prices_4064601.sql
+++ b/cowprotocol/accounting/slippage/slippage_prices_4064601.sql
@@ -25,21 +25,49 @@ with token_times as (
     group by 1, 2
 ),
 
+aave_tokens_mapping as (
+    select distinct
+        chain as blockchain,
+        atoken as atoken_address,
+        asset as underlying_address
+    from aave_v3_multichain.poolconfigurator_evt_reserveinitialized
+    union distinct
+    select distinct
+        'ethereum' as blockchain,
+        atoken as atoken_address,
+        asset as underlying_address
+    from aave_v3_lido_ethereum.poolconfigurator_evt_reserveinitialized
+    union distinct
+    select distinct
+        'ethereum' as blockchain,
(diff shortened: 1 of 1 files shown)


cowprotocol/accounting/slippage/slippage_prices_4064601.sql
```diff
@@ -25,21 +25,49 @@ with token_times as (
...
+    from aave_v3_lido_ethereum.poolconfigurator_evt_reserveinitialized
+    union distinct
+    select distinct
+        'ethereum' as blockchain,
+        atoken as atoken_address,
+        asset as underlying_address
+    from aave_v3_ethereum.etherfipoolconfiguratorinstance_evt_reserveinitialized
+),
+
+token_times_enriched as (
```
fhenneke, line 48: I would probably add a comment here of the form
```
-- All tokens are associated with a proxy address which is used for finding a price feed instead of the original contract address
```

fhenneke (approved):
Change looks reasonable.

I did not test anything except for the fact that the query does not crash.

PR #251, opened 2025-09-16
Exclude txs from slippage for Sep9-16, 2025

harisang:
This PR excludes 3 txs from slippage on mainnet. They all include the same token that does some additional transfers from the Vault Relayer to the settlement that is perceived as slippage

https://etherscan.io/tx/0x149c90ed4aa005a8aefc4b8bb85bfeaf31f35151989953a7cac3187ae8e9cb2d

https://etherscan.io/tx/0xc3e9ab2accbc9ba20868a55df63374b81edc8d3e92b9528d61085792267f744b

https://etherscan.io/tx/0xf15e37449c1492b18088e42608a281430a8ad45feba5a7763c8e587582ac1dd9


fhenneke (approved):
Slippage for these transactions definitely looks strange.

Not sure what that token does. Maybe it should be disallowed.

PR #318, opened 2026-01-15
Add base query for analysing CIP 74

fhenneke:
This is a draft PR for adding a base query on analysing the impact of CIP 74 on solvers.

The goal of the query is provide data per auction and solver on current and alternative rewards (e.g. the old reward scheme before CIP-74, changed volume fees).

The query is based on `rewards_per_auction_5787562.sql`. Main changes are
* A new column `xrate_type` classifying batches into those containing only trades on token pairs with `'stable'` (i.e. stongly correlated) and `'variable'` exchange rate. It is based on query 5719467 (see also #317) and its use in query 6481803.
* A new protocol fee computed as `protocol_fee + volume * (if(xrate_type = 'stable', {{volume_fee_bps_stable}}, {{volume_fee_bps_variable}}) - 2.0) / 1e4`. It removes the 2 bps fee and adds a fee depending on the classification in `'stable'` and not `'stable'`.
* Capping parameters are hardcoded to allow for computing old rewards.
* Some parameters of the old query are removed (e.g. on fixed fees or price improvement).

One way to approach for checking the query is to look at the changes compared to the query it is based on via 
```
git diff --no-index --color ../../accounting/rewards/reward_capping_experiments/rewards_per_auction_5787562.sql rewards_per_auction_6517101.sql
```
(from the folder of the new file).

### DIFF
--- a/cowprotocol/mechanism/cip_74_analysis/rewards_per_auction_6517101.sql
+++ b/cowprotocol/mechanism/cip_74_analysis/rewards_per_auction_6517101.sql
@@ -71,6 +71,7 @@ aggregated_batch_data as (
         rbd.auction_id,
         rbd.solver,
         rbd.uncapped_payment_native_token as uncapped_reward,
+        sum(rbd.winning_score) as score,
         coalesce(sum((rod.protocol_fee - coalesce(rod.partner_fee, 0)) * rod.protocol_fee_native_price), 0) as protocol_fee, -- this is the actual revenue of the protocol
         coalesce(sum(case when t.order_type = 'SELL' then t.atoms_bought * rod.protocol_fee_native_price else t.atoms_sold * rod.protocol_fee_native_price end), 0) as volume,
         bool_and(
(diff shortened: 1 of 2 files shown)


cowprotocol/mechanism/cip_74_analysis/rewards_per_auction_6517101.sql
```diff
@@ -0,0 +1,130 @@
...
+        xrate_type,
+        protocol_fee,
+        volume,
+        protocol_fee + volume * (if(xrate_type = 'stable', {{volume_fee_bps_stable}}, {{volume_fee_bps_variable}}) - 2.0) / 1e4 as new_protocol_fee,
+        uncapped_reward
+    from aggregated_batch_data
+    where
+        (
+            native_prices_are_accurate -- this filters out cases where native prices are not accurate
+            and abs(uncapped_reward) < volume -- this filters out cases where native prices were corrected but rewards are too large
```
harisang: this would remove cases where a solver had 2 winning solutions, one with very small volume that got settled and the other one with significant volume and surplus that reverted, right?

fhenneke: I reverted to what I used in the original query.

data-cowwboy (approved):

PR #341, opened 2026-03-16
Exclude transaction from Base

bram-vdberg:
# Description
This PR excludes the following transaction from the Base payouts: https://explorer.cow.fi/base/tx/0x7FBFA52D0B86756A6C748A0BDE34E58B8CEC50C6EF46B260C42A7F8BAD3B98A1?tab=orders

# Context
Due to a bug one of the solvers erroneously filled a partially fillable order with the full amount, leading to an excessive surplus. Since this transaction occurred between two DAO safes no actual unusual slippage occurred so this PR will prevent us from deducting slippage from the solver's payouts.


bram-vdberg:
Closing because payouts the tx is too old to be relevant for any current payouts

PR #355, opened 2026-05-25
Add bogus_native_prices alert query (sell orders)

data-cowwboy:
## Summary
- Adds the SQL for Dune query [`7574432`](https://dune.com/queries/7574432) under `cowprotocol/alerts/`, matching the existing `solver_inactivity_alert_6568090.sql` naming convention.
- The query flags settled SELL orders where the sell-side and buy-side native values diverge by >1000x (likely bogus native price), with a `min_native_volume` floor so dust orders don't trip it.
- Consumed daily by the `bogus_native_prices` monitor in cow-dagster (companion PR: cowprotocol/cow-dagster#nitish/bogus-native-prices-alert).


cowprotocol/alerts/bogus_native_prices_alert_7574432.sql
```diff
@@ -0,0 +1,68 @@
...
+    case
+        when sell_value_native > 1000 * buy_value_native then 'sell_side_overpriced'
+        when buy_value_native > 1000 * sell_value_native then 'buy_side_overpriced'
+    end as anomaly_type
+from sell_orders
+where (
+    sell_value_native > 1000 * buy_value_native
+    or buy_value_native > 1000 * sell_value_native
+)
+and greatest(sell_value_native, buy_value_native) > {{min_native_volume}}
```
harisang: the left-hand side should be minimum/least or whatever dune uses, in my opinion, as you could have cases where i am selling 20 dollars on mainnet and most of which is consumed by gas fees.

cowprotocol/alerts/bogus_native_prices_alert_7574432.sql
```diff
@@ -0,0 +1,68 @@
...
+        t.order_type = 'SELL'
+        and t.block_time >= timestamp '{{start_time}}'
+        and t.block_time < timestamp '{{end_time}}'
+        and rod.quote_sell_token_price is not null
+        and rod.protocol_fee_native_price is not null
+        and rod.quote_sell_token_price > 0
+        and rod.protocol_fee_native_price > 0
+)
+
+select
```
harisang, line 48: We should add auction_id here as well, and environment and blockchain.

harisang (changes requested):
added some comments

harisang (approved):

PR #371, opened 2026-07-14
Un-vouched solvers approval query

tamir-cow:
Created new query that checks token approvals from the settlement contract that were signed by recognized solvers whose addresses lack current bonding pool vouches—identifying potential risk from unverified solver signings.


harisang:
Can we maybe invoke the existing query that we have that detects approvals instead of rewriting the code?

E.g., this is how invoke a query within the slippage query (completely irrelevant queries, but just to showcase the syntax
```
select * from 
"query_4059683(blockchain='{{blockchain}}',start_time='{{start_time}}',end_time='{{end_time}}',raw_slippage_table_name='raw_slippage_breakdown')"
```

harisang:
Looks good. I think we should also look for solvers that are not whitelisted anymore. As there can be cases where a solver has been blacklisted but somehow was not unvouched. Technically the bonding pool then still covers you but in practice we don't want to be going after deprecated solvers, plus the vouching might still be active but the bonding pool might not be there anymore

harisang:
Applied the proposed changes to the 2 queries and things look good! Approving, but would suggest to replace the term "deprecated" for "blacklisted" so as to be more explicit

harisang (approved):

